Issue #68, opened 2018-04-11
Improve rate limiting


yonjah:
I was going over the security wiki https://github.com/nodemailer/wildduck/wiki/Security-implementation
and saw that rate limiting is done per user account.
Reading the code it seem like limiting is only happening if the user exists.

I can think of two issues that this might cause -
User enumeration - you can send a login request multiple time and if you hit the rate limiting you know the user actually exist.
Denial of service - you can keep sending fake login attempts to the server preventing the user from logging in.

I think this can be improve by first limiting auth by ip address as well so a single ip can only do x login attempts. if x is much smaller that he rate limit on the user account it will make both attacks a harder requiring sending requests from multiple ips to reach the rate limit of a single account.

andris9:
I changed rate limiting in 97af5287e49a083186bfc75faa1576c0439a0361 as such:

  * non existent users are also rate limited, so it would not be possible to enumerate users
  * IP address is used as part of the rate limit key, so rate limiting applies to the same username:ip pair, not just to a username

I have mostly seen two kind of issues with rate limiting:

  1. Denial of service, where entire account blocks for a long time after x-failed attempts. Someone intentionally used invalid password to lock an user out of his email account. This is the reason why I decreased the lock time window, to a small length (60 seconds counting from first invalid attempt) as it would prevent brute force attacks while not locking out the user for too long time. It does not prevent malicious denial of service though, so now I added the IP check as well
  2. Slow password crawling from a large array of IP addresses doing limited authentication attempts over a long period like half a year. This is not mitigated from current rate limiting system at all (invalid authentication attempts are quite rare)

andris9:
Rate limitation changes are now published

PR #69, opened 2018-04-13
rate limiting by ip (relates to #68)

yonjah:
This is just a POC I didn't ran the code or any tests so this should be taken with a grain of salt.

limiting is done both by user id and ip address but in two separate buckets.
The issue with using a single bucket is that a single IP can try many users with no limit
And many IPs can target a single user with no limit. both are common scenarios.
So by separating them we solve this two issues as well.

I changed the Auth window a bit. my thought is that one malicious IP will not be able to do a DOS on a specific user. and having a two minutes is still small enough for the user not being too affected, yet a legitimate user trying to login from a single ip will have 10 attempts per 5 mins which is probably enough for any legitimate use.

IP based limits are not removed on successful login

I didn't touch `authenticateUsingOldPassword` I grep the code base and couldn't find it being referenced anywhere and I think there might be more issues to consider with this method.

Indentation is a bit off but I wanted the diff to show changes clearly.

relates to #68

### DIFF
--- a/lib/consts.js
+++ b/lib/consts.js
@@ -38,9 +38,14 @@ module.exports = {
     BCRYPT_ROUNDS: 12,
 
     // how many authentication failures per user to allow before blocking until the end of the auth window
-    AUTH_FAILURES: 6,
+    USER_AUTH_FAILURES: 12,
     // authentication window in seconds, starts counting from first invalid authentication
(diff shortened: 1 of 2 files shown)


andris9:
Seems great, thanks!

PR #84, opened 2018-06-08
Breaking up install.sh into chapters

bunyevacz:
Fixed two issues:
82: need to install npm globally to get around permission problems
83: need to start nginx service before restarting it

Better executable requirements (lsof, ps). It may be not installed
on minimal systems.

Better service detection on given port. It is especially useful,
if the installation.sh got interrupted for some reason, and
already installed some services.

Minor doc update, so a single line is required to paste in terminal.
curl vs. wget -> stayed with wget, it is installed by default on ubuntu

Colors: added color support for the terminal output:)


andris9:
Nice, thanks, seems a lot better now!

PR #88, opened 2018-06-08
DKIM and SPF correction, some renaming

bunyevacz:
I renamed some install scripts to be more clear.

SPF: Suggest [MAILDOMAIN], [HOSTNAME] and [IP ADDRESS]
Its a better practice to be more inclusive when it comes
to dns SPF records.

DKIM: Some dns registrars truncate dns TXT records at
255 chars. So 2048bit do not fit (about 390 vs. 230 chars).
So 1024bit keys are a good choice, after all it is
only a mail verification mechanism, do not encrypt the mail...

Show tip how to stop systemd service (03_install_check_running_services.sh)

sudo su prefered to become root, sudo su fails with npm permission errors
when installing (dunno why).

A mini tutorial is shown at the end about SPF, DKIM and
how to add/remove/modify DKIM keys.

This pull request closes issue 85,86.


andris9:
Thanks!

PR #102, opened 2018-09-02
fixed apidocs for "Users - GetUser"

erdii:
Pretty self-explanatory.

Some sections were missing under "limits".


andris9:
Thanks!

PR #119, opened 2019-03-02
Refactoring of address resolution

rqelibari:
In order to extend the address resolution easily with features like:

- Partial wildcard address recognition (e.g. *prefix@...)

or

- Multiple domain aliases

we refactored the `lib/user-handler.js` module and made the `resolveAddress` a mixin in its own file.
Instead of multi stack closures we use a simple context object. What do you think @andris9?


andris9:
Sorry for the delay, somehow missed this PR.

Proposed idea is good and I would add it to WildDuck. Refactoring is to a different direction though than I've gone. I've stopped using callbacks wherever possible and have moved to async..await. I have rewritten large parts of WildDuck as well to use async..await, for example see the API [handlers](https://github.com/nodemailer/wildduck/blob/master/lib/api/addresses.js). So for now I would ask you to either use the old callback syntax or if you want to help with refactoring, then use async..await as in the other parts of the code. Introducing new concepts would only make it more difficult to refactor it again in the future.

To make things easier as some methods that you need to use but are quite big and difficult to convert, I have usually converted these methods to Promises with `util.promisify()`:

```
const putMessage = util.promisify(messageHandler.put.bind(messageHandler));
...
let resp = await putMessage(args);
```

rqelibari:
Code refactoring is postponed. Partial wildcard recognition is implemented with PR #127.

PR #127, opened 2019-05-10
Allow (forward-)partial wildcard addresses

rqelibari:
**Description**
This PR implements recognition of forward-partial wildcard addresses à la
`*_somestring@domain.com`. This means the user can receive mail for any address ending in `..._somestring@domain.com` by just adding the wildcard address `*_somestring@domain.com` to his account.

This was originally implemented in #119, which mixed the feature with further code refactoring.


andris9:
Did some modifications regarding address matching:
* up to 32 characters of the wildcard suffix are checked
* longest match is preferred so "test@domain" always matches "*est@domain" even if there is "*st@domain"

PR #140, opened 2019-07-08
an user > a user

louis-lau:
Heya,

English is a weird language. And it's actually "a user", even though it starts with a vowel.

> "An" goes before all words that begin with vowels:
> 
> An egg
> With two exceptions:
> 
> When "u" makes the same sound as the "y" in you, or "o" makes the same sound as "w" in won, then "a" is used:
> 
> a union
> a united front
> a unicorn
> a used napkin
> a U.S. ship
> a one-legged man

All changes I made here should be purely comments and documentation, let me know if that's not the case.


andris9:
Thanks a lot! Somehow never knew about this

PR #145, opened 2019-08-16
404 code to 200 code with an error code string

louis-lau:
404 error without an error code string is inconsistent when compared to all other methods. While on that topic, is there any reason for not using the right http response codes?

Sorry for all the issues 😁. Knowing me there's more to come. I'll get on the SHA password encryption as soon as I get my application working enough.

### DIFF
--- a/lib/api/filters.js
+++ b/lib/api/filters.js
@@ -475,9 +475,9 @@ module.exports = (db, server) => {
             }
 
             if (!r.deletedCount) {
-                res.status(404);
                 res.json({
-                    error: 'Filter was not found'
+                    error: 'Filter was not found',
+                    code: 'FilterNotFound'
                 });
                 return next();
             }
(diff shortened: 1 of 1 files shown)


andris9:
At first WIldDuck API always returned HTTP 200 with either `{success: true,...}` or `{error:"..."}`. Then I started using the `code` property for errors (mostly because of client translations) and then also started using different HTTP codes than 200 for errors. Both of these additions are not completed as I've only added these whenever I've worked on some part of the system. Kind of the same way some parts of the code use callbacks, some part use async..await. Whenever I work on something that uses callbacks I try to change it to use await instead.

So regarding this PR I'd probably keep the 404 response code.

louis-lau:
Ah, makes sense. I added back the 404. I hope adding 404 responses to other methods in the future won't break any clients handling the errors.

I still feel the error code string is important in this case, as a 404 could mean account not found or filter not found.

PR #162, opened 2019-11-28
Migrated from joi to @hapi/joi

### LINKED ISSUE #161: Moving from Restify to Hapi
The package `joi` seems to be depreciated according to it's [npm page](https://www.npmjs.com/package/joi). The author recommends switching to `@hapi/joi`.

I'll put up a PR to fix this soon

rakshith-ravi:
Fixes #161 

If I'm being completely honest, I have no experience in using `joi` and migrating it was a little difficult, considering the documentation for the package is sparse.

However, with the current set of changes, `npm run test` runs, but fails in 3 specific cases. Namely:
- `IMAP Protocol integration tests "before each" hook for "should list capabilites"`
- `API tests "before all" hook for "should POST /domainaliases"`
- `Send multiple messages "before all" hook`

I don't think they're related to this library, but I'm not sure. Would appreciate it if someone can take a look at this and let me know what exactly is happening. If you're able to fix this up, I would really appreciate it if you can also let me know what exactly went wrong.


andris9:
Fixed with 488f41c93eb7fc18d144d8b1b02e27bb70a3256f

Issue #79, opened 2018-05-14
install.sh vs ubuntu 18.04


ghost:
More of an info than an issue. Had to do the following to get the install script running on ubuntu 18.04

```
root@shinymail:~# diff -ur install.sh install.sh.ubuntu.18.04 
--- install.sh	2018-05-14 15:47:01.219007025 +0200
+++ install.sh.ubuntu.18.04	2018-05-14 15:47:37.367388180 +0200
@@ -100,9 +100,7 @@
 chown -R deploy:deploy /home/deploy
 
 # mongo
-apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 0C49F3730359A14518585931BC711F9BA15703C6
-gpg --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58712A2291FA4AD5
-gpg --armor --export 58712A2291FA4AD5 | apt-key add -
+apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2930ADAE8CAF5059EE73BB4B58712A2291FA4AD5
 echo "deb [ arch=amd64,arm64 ] http://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/3.6 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-3.6.list
 
 apt-get update
root@shinymail:~# 
```

Falieson:
Also, on GCP (not on DO) in 18.04 I had to install these packages to get the script running
```
apt install -y lsof dirmngr
```

andris9:
The current install scripts works in Ubuntu 16.04 and 18.04 in DO

Issue #71, opened 2018-04-28
Docker Image


AddoSolutions:
Hey guys, just wanted to see if there was anyone already working on a docker image for this project.  I have a pretty long history with Docker, and have dockerized a few other projects so I can certainly help significantly with this if need be.

The one really big help I have found is your install script.  It doesn't appear to be fully documented, but it only seems to work on Ubuntu Xenial flavor?  With that image, I was able to get it to install fairly well into an image, but there are several important tweaks for the install that would increase it's stability/standardize it's installation.

This is a VERY cool project, and glad that someone is FINALLY working on a replacement for the horrid file based mail systems written in a modern language.  I actually started doing it myself but ran out of time to finish (some of my progress is [here](https://github.com/AddoSolutions/unsuckish-mail) if it's of any help, though this repo is FAR more advanced).  

Let me know if I can help on this project.  I am pretty seasoned in NodeJS, MongoDB, and Docker -Redis being my weak point.

Nick

AddoSolutions:
For anyone wanting to play with docker, you can get it about as far as I did with:
do:

```shell
# On the host
docker run --rm -it --link mongodb --link redis -p 8080:8080 -p 993:993 ubuntu:xenial bash

# In the newly created temporary container
apt-get update; \
    apt-get -y install lsof wget gnupg dnsutils curl sudo git; \
    wget https://raw.githubusercontent.com/nodemailer/wildduck/master/setup/install.sh; \
    chmod +x install.sh; \
    ./install.sh ; \
    echo 'mongo="mongodb://mongodb:27017/wildduck"
    redis="redis://redis:6379/3"' > /etc/wildduck/dbs.toml; \
    cd /opt/wildduck; \
    /usr/bin/node server.js --config="/etc/wildduck/wildduck.toml"
```

andris9:
Actually there is nothing Ubuntu specific, for example I run the server in a custom CoreOS based system. The install script uses apt for installing the dependencies (MongoDB and such), and it uses systemd to run the services (thus the requirement for Ubuntu 16, not 14 which does not have systemd). If you manage Ubuntu and Redis separately and start the service process as a docker command then there's nothing you'd need from Ubuntu.

AddoSolutions:
Very cool! Yeah it does a really nice Job, I’m just trying to figure out how to best dockerize it. 

From a strategic perspective is Docker something you guys are interested in formally supporting or am I going to be better off doing it separately?

andris9:
[...]

PR #180, opened 2020-02-14
Dockerize the repo

rakshith-ravi:
Worked on by @thebongy

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+node_modules
+Dockerfile*
+docker-compose*.yml
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM node:lts-alpine
+
+RUN apk add --no-cache make git dumb-init python
+
+WORKDIR /wildduck
+COPY . .
+
+RUN npm install --production
+
+ENV WILDDUCK_APPDIR=/wildduck \
+    WILDDUCK_CONFIG=/wildduck/config/default.toml
+
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+CMD node ${WILDDUCK_APPDIR}/server.js --config=${WILDDUCK_CONFIG}
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,8 @@ COPY . .
 RUN npm install --production
 
 ENV WILDDUCK_APPDIR=/wildduck \
-    WILDDUCK_CONFIG=/wildduck/config/default.toml
+    WILDDUCK_CONFIG=/wildduck/config/default.toml \
+    CMD_ARGS=""
 
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
-CMD node ${WILDDUCK_APPDIR}/server.js --config=${WILDDUCK_CONFIG}
+CMD node ${WILDDUCK_APPDIR}/server.js --config=${WILDDUCK_CONFIG} ${CMD_ARGS}
--- a/README.md
+++ b/README.md
@@ -96,6 +96,30 @@ See [API Docs](https://api.wildduck.email/#api-Users-PostUser) for details about
 
 Any IMAP or POP3 client will do. Use the credentials from step 4\. to log in.
 
+### Docker Install
+The easiest way to setup wildduck with a docker image is given below, for more documentation about configuration options in the docker image, refer to
+the [wiki page on the Docker](https://github.com/nodemailer/wildduck/wiki/Docker).
+
+
+A docker hub image built using the [Dockerfile](./Dockerfile) in the repo is also available
+
+To pull the latest pre-built image of wildduck:
+
+```
+docker pull nodemailer/wildduck
+```
+
+It is also possible to pull a specific version of wildduck by specifying the version as the image tag.
+(example, for version 1.20):
+```
+docker pull nodemailer/wildduck:1.20
+```
+To run the docker image using the [default config](./config/default.toml), and `mongodb` and `redis` from the host machine, use:
+```
+docker run --network=host nodemailer/wildduck
+```
+
+
 ## Goals of the Project
 
 1.  Build a scalable and distributed IMAP/POP3 server that uses clustered database instead of single machine file system as mail store
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,30 @@
+version: "3.7"
+services:
+    wildduck:
+        network_mode: host
+        build: .
+        ports:
+            - "8080:8080"
+            - "143:143"
+            - "110:110"
+            - "993:993"
+            - "995:995"
+        depends_on:
+            - redis
+            - mongo
+
+    redis:
+        network_mode: host
+        image: redis:alpine
+        volumes:
+            - /data
+        expose:
+            - 6379
+    mongo:
+        network_mode: host
+        image: mongo
+        restart: always
+        volumes:
+            - /data/db
+        ports:
+            - 27017


rqelibari:
Hi @rakshith-ravi,

thank you for your efforts to dockerize Wildduck.

Currently your approach is just a (dockerized) wrapper around the install scripts that does  does not integrate any sophisticated feature, that would make sense, when using a repository-integrated docker solution. @AddoSolutions already proposed a simple container like your's (see [here](https://github.com/nodemailer/wildduck/issues/71#issuecomment-385190084)) in issue #79.

There has been a great discussion over at issue #71 on what features a repository-integrated docker solution should offer. We, at Astzweig, took another approach at [astzweig/docker-wildduck](/astzweig/docker-wildduck), and are currently missing the resources to implement those additional features. Maybe you can use or repository as a base and try to develop solutions for the missing features.

rakshith-ravi:
@rqelibari, our goal is to setup separate docker images for each parts of wildduck, while providing a docker-compose.yml file that acts as an example. This way, each part of wildduck can scale independently as a separate container. If these goals align with yours, we'd be happy to work on your image. That being said, we'd also prefer to have it integrated with this repo and built using travis to have a consistent version number across both the non-docker setup and the docker version. Does that sound like something you'd be willing to work on? If that sounds good to you, we'd be happy to contribute to your repo instead

rqelibari:
@rakshith-ravi Thank you for the clarification. That sounds like a good approach in my opinion, as it produces many small containers and people can decide themselves, what components of the wildduck ecosystem they want to use (e.g. haraka as inbound SMTP or something else).
Tying the versioning of each container to the specific service makes sense, too, using this approach. In my opinion it is the only approach that can be included in the repository directly.

## Approach of Astzweig
The repo at [astzweig/docker-wildduck](/astzweig/docker-wildduck) on the other side is about providing a fully functional email server based on wildduck. In such a container tying the version to one single service is therefor not viable.

rakshith-ravi:
Thank you for the feedback. Much appreciated.

As far as the PR goes, could you please help us review it? We're trying to understand if there are any issues in it so far or if we are even working on the right direction. I've tried reaching out to @andris9 (both here and on gitter), but I'm not able to get through to him. Could you please review this PR and let us know if it's fine?

andris9:
@rakshith-ravi  @rqelibari I don't really have an opinion of how docker image should look like. The problem is that it can be composed in so many different ways and there's always someone who wants to do it some other way, this is why I haven't merged so far any of the dockerization PRs. If anyone has asked, I've pointed to the Astzweig repo.

If you can somehow figure out together the best way of how docker setup should look like, I'm willing to merge it to the repo.

rakshith-ravi:
I personally think we should have multiple docker images for each service that wildduck offers, and bundle them all together in a compose file that can be used as an example (for someone who's just all about "show me the code to get it up and running"). This would also give the option of decoupling the services and scaling them up individually if required.

That being said, I think we can keep going on and on about the right way to do it, when in reality, not a lot of decisions make _that_ much of a difference. As long as it doesn't affect scalability and simplicity of use (which, if I'm not wrong, are some of the main goals of this project), we can go with any way of doing it. We propose the above to make things easier. We would really like some opinions from other people on the shortcomings of our approach and how we can improve this. If no one has any issues, I'd suggest we merge this and take a look at it later on when we face any issues with it

rqelibari:
@andris9 That captures the situation quite well!

@rakshith-ravi Exactly. I think the idea to provide a docker container for each part of wildduck makes sense. It should be easy for users though, to use a specific container, e.g. users should be able to use ready-made containers from [Docker Hub](https://hub.docker.com) and run them without too much trouble.

### The configuration problem
Wildduck needs to know the address of the `mongo` server to run properly,  which can be configured at [db.toml](/nodemailer/wildduck/blob/master/config/dbs.toml#L2). So the user will most likely need to change this configuration option. To change that configuration option though, the user needs to have the whole source code of the [wildduck repository](/nodemailer/wildduck) to rebuild the entire container. A Docker Hub instance of the container would make not sense. Hence a runtime configuration seems to be a better approach. The [12 factor container](https://medium.com/notbinary/the-twelve-factor-container-8d1edc2a49d4) article of notbinary has a good explanation on that topic.

### The environment solution
I suggest, making the container configurable by environment variables. At least the most important configuration options (see the list over at [astzweig/docker-wildduck](/astzweig/docker-wildduck#environment-variables) for examples).

#### Cocof
At [astzweig/docker-wildduck](/astzweig/docker-wildduck) we developed a python package called `cocof` ([cocof at PyPi](https://pypi.org/project/cocof/)) to configure the container at runtime. E.g. to configure wildduck, you could run the followin line in a shell script:

```bash
cocof '/wildduck/config/default.toml' '[{"op": "add", "path": "/emailDomain", "value": "example.com"}]';
```
You can take a look at the script [_wildduck.sh](/astzweig/docker-wildduck/blob/master/scripts/_wildduck.sh) to see how we configure wildduck at runtime.

rakshith-ravi:
How about mounting the db.toml file on an external volume and changing it's contents accordingly?

louis-lau:
This is fine if you only have one server. In a multi server setup with docker swarm or kubernetes you don't want to store the config in volumes. It would make it harder to scale.

rakshith-ravi:
Can't the volumes be shared across the containers?

But yeah, I do see your point. It's not the most scalable. But wouldn't the same issue be faced in the non-dockerized version of wildduck as well?

louis-lau:
> Can't the volumes be shared across the containers?

Sure, but a volume is bound to the host the container is running on. This can change, and you need the config on all running hosts.

> But wouldn't the same issue be faced in the non-dockerized version of wildduck as well?

Yes! Which is why a dockerized version would be great :)

rakshith-ravi:
> Yes! Which is why a dockerized version would be great :)  

Haha yeah point taken. 

So I was thinking that the config files can be shared across hosts using an NFS mount. Since they're not gonna be accessed often, the speed of access doesn't really matter. But other than that, it'll be very difficult to change the way that works without modifying the way wildduck loads config data. I'm not sure if @andris9 is willing to change that.

louis-lau:
Andris doesn't have to do it, it could be a PR if you had the time. The problem I see with that is that environment variables have character restrictions, and their contents can only be a string.

You could get the environment variables and transform them to a toml/json/js config when starting the container.

thebongy:
> This is fine if you only have one server. In a multi server setup with docker swarm or kubernetes you don't want to store the config in volumes. It would make it harder to scale. Environment variables are the way to go.

For docker swarms, instead of using a volume, docker has [`docker config`](https://docs.docker.com/engine/swarm/configs/), which basically shares configuration files across multiple nodes in the swarms. This lets you mount the config files, in this case `default.toml` across multiple nodes, just like you would mount a volume

For kubernetes, there is a similar solution with [`configmap`](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/)

I think for a use case involving swarms/kubernetes, these options are much easier to configure (and cleaner) rather than having to modify multiple environment variables, which would be hard to maintain as the configuration options change/new configuration options are introduced. This also lets us decouple configuration from the actual container, thus keeping the containers portable, which is one of the main principles of containerization

louis-lau:
Damn, thanks @thebongy!

rqelibari:
@thebongy Good point about the problem of keeping environment variables aligned with their corresponding configuration option. Still, talking from a UX position, I think there should be environment variables for the most important configuration options - options that will most probably need to be changed by every user -  like the FQDN.

### UX Reasoning
If probably everyone has to tell wildduck their servers FQDN than a simple environment variable is easier defined than using `docker config` to first define a configuration file, than finding out, where the correct path is and so on.

### The mysql container
If you take a look at the official [mysql container](https://hub.docker.com/_/mysql) you can see, that they take the same approach. Everyone is most probably going to set their own `mysql` root password, so despite of [`docker secrets`](https://docs.docker.com/engine/reference/commandline/secret/) they offer a environment variable to configure the `mysql` database that way, simply as a UX decision.

thebongy:
@rqelibari I've had a look at the repo you mentioned [astzweig/docker-wildduck](https://github.com/astzweig/docker-wildduck), it provides a really nice way to setup common configuration options across Wildduck, zonmeta, and haraka, but the goal here is to setup a Dockerfile for wildduck as independent of other repos as possible.

I totally get the UX argument, but I don't think it is necessary to have a script in the Dockerfile to write env variables to the config files. There is another alternative to use [wild-config](https://github.com/nodemailer/wild-config) which already has a feature to pass configuration through command line options.

So I could provide a provision in the Dockerfile on this  PR to pass command line args to the RUN command as such:

```
ENV CMD_ARG
RUN node server.js ${CMD_ARGS}
```

and let's say a user wanted to configure the mongo and redis URL, he could pass them to docker run as env args, instead of having to mount the config file as volume or use `docker config`.
```
docker run -e CMD_ARGS="--dbs.mongo='mongo://mongo_host' --dbs.redis='redis://redis_host'" wildduck
```

This approach has the benefit of again, keeping the container image as independent of the configuration as possible, thus making it easier to maintain in the future.

I think along with some documentation on the wildduck README on how to use the `CMD_ARG` option to set common fields like the `FQDN` , this can work better, from a UX perspective. What do you think?

rqelibari:
@thebongy That is a good idea using a wildcard environment variable to pass options to `wildduck` in combination with the documentation of the most important options in the README file.

Hence, in order to merge this PR we need two commits:

- [ ] 1. Change the Dockerfile as described by @thebongy.
- [ ] 2. Add a section in the README file that either describes the usage of `CMD_ARGS` or links to an article in the Wiki. In my opinion at least the following configuration options should be described:
  - Setting the FQDN of the server.
  - Setting the default domain name for which the email server shall receive emails.
  - Setting the domain name of the `redis` host.
  - Setting the connection link of the `mongo` host.

Everyone all right with that?

### Docker Hub
@andris9 In order for us to publish a Docker container of wildduck on Docker HUB, you will need to register on Docker HUB ([here](https://hub.docker.com)). Afterwards I can make you the owner of the `nodemailer` Organization, that I created on Docker Hub, so that you can link either your own account or a specific service account to Docker Hub ([Docker Hub service accounts](https://docs.docker.com/docker-hub/builds/#service-users-for-team-autobuilds)).

andris9:
@rqelibari my docker hub username is andris9

thebongy:
@rqelibari the Dockerfile now accepts the env variable `CMD_ARGS`. Have included a quick way to get the docker container running on the repo's `README.md`.

Additionally, I have created a wiki page @ (https://github.com/nodemailer/wildduck/wiki/Docker) with more detailed instructions to configure the container

rqelibari:
@thebongy That is awesome, nice work!
@andris9 You can merge this PR now. Afterwards you need [to link](https://hub.docker.com/orgs/nodemailer/settings/linked-accounts) either your own github account with the nodemailer organization or create a "service" account, as described [here](https://docs.docker.com/docker-hub/builds/#service-users-for-team-autobuilds) and link that one, so that I can configure autobuild.

andris9:
Looks great indeed!

rakshith-ravi:
Is the docker image not published to docker hub?

andris9:
@rakshith-ravi @rqelibari  I forgot to link Github account with Docker. Did this just now, so wildduck repo should be accessible from Docker Hub.

rakshith-ravi:
Thank you @andris9. @rqelibari can we setup autobuilds with the correct tags please?

rqelibari:
Docker Hub's autobuild is set up and the first build is finished: [https://hub.docker.com/r/nodemailer/wildduck](https://hub.docker.com/r/nodemailer/wildduck)

Every change on the `master` branch will trigger a new build of the `latest` container. Every new tag will trigger a new build of the container with the same tag.

rakshith-ravi:
Awesome! Thank you @rqelibari! 

Is there any way we can trigger builds for existing tags as well?

rqelibari:
> Is there any way we can trigger builds for existing tags as well?

@rakshith-ravi No, as the repository did not have a Dockerfile before.

rakshith-ravi:
> > Is there any way we can trigger builds for existing tags as well?
> 
> @rakshith-ravi No, as the repository did not have a Dockerfile before.

@rqelibari it looks like the newer tags that were published aren't updating on docker hub either. Can you please see if something is configured wrong?

rqelibari:
@rakshith-ravi You are right, thanks for noticing. I had missed a 'v' in the regular expression. Just reconfigured it.

Before:
Screen Shot 2020-04-06 at 21 08 55

After:
Screen Shot 2020-04-06 at 21 19 49

rakshith-ravi:
Got it. Thanks

rakshith-ravi:
Does this mean the tags will appear in docker hub?

https://hub.docker.com/r/nodemailer/wildduck/tags

I'm not able to see the tags yet. I'll check again tomorrow.

rqelibari:
> Does this mean the tags will appear in docker hub?

@rakshith-ravi No, only new tags will appear in docker hub. So the next time andris tags a commit, docker hub should create a tagged build.

PR #199, opened 2020-04-18
Update 04_install_import_keys.sh

halfengineer07:
remove sudo for docker file


halfengineer07:
remove sudo so can be run in dockerfile

PR #210, opened 2020-05-08
Allow disabling forwarder

louis-lau:
Allow disabling a forwarded address, just like you would disable a user. This results in a `550 Mailbox disabled, not accepting messages`.

I used the variable name forwardedDisabled as I thought it would reduce confusion when listing normal addresses together with forwarded addresses. Let me know if you have other thoughts.

A PR on haraka-plugin-wildduck should be linked below 😁


andris9:
Seems great

PR #213, opened 2020-05-14
Added link to Project Pending

andris9:
Added link to Project Pending to 3rd party software docs.


andris9:
@louis-lau how did you think the ordering of these projects should be organised? Does the "by commit date" mean that older projects go first?

louis-lau:
Yeah that my intention, would get too tempting to put my own project on top otherwise 😁. But do whatever you want, maybe they should be sorted by relevance, though I find that hard to quantify.

andris9:
Ok, it's totally fine. I was adding one of my own projects and did not want to give an advantage to it, better treat everyone equally on that side.

louis-lau:
😁

louis-lau:
I like the cat detective idea

PR #243, opened 2020-07-13
Docker autopublish

figassis:
This PR implements basic cicd using gh actions provided the DOCKER_USERNAME, DOCKER_PASSWORD and IMAGE_NAME secrets are available to the repository. A push to master or develop will trigger a docker build and push to docker hub using the version in package.json and the current unix timestamp in minutes, like 1.26.4-26577799.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -1,0 +1,26 @@
+name: Docker Image CI
+
+on:
+  push:
+    branches:
+      - master
+      - develop
+  pull_request:
+    branches:
+      - master
+      - develop
+
+jobs:
+
+  build:
+ 
+    runs-on: ubuntu-latest
+ 
+    steps:
+    - name: Checkout Current branch
+      uses: actions/checkout@v1
+    - name: Build Docker Image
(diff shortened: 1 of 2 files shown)


louis-lau:
What's the difference between this and docker hub auto building when a new tag is set? Because that is already happening right now.

figassis:
> What's the difference between this and docker hub auto building when a new tag is set? Because that is already happening right now.

This is mostly for testing/debugging, my use case is I deploy the images on a k8s cluster, and usually I might test multiple changes on a test deployment before deploying to production, and k8s/helm will cache images, so in a way this is a cache busting solution. It won't trigger if the secrets aren't in the repo, and also won't return an error, it will just exit, so its optional for everyone sharing this use case.

andris9:
Is there any way to do the same thing without including the code in git repo? I've generally tried to keep Docker out of WildDuck because there are unlimited ways of doing things and whatever you do - someone needs to do it in some other way.

figassis:
You're right, I was trying to keep maximum compatibility with the upstream repo, but yes, these don't need to be there, this pr can be dropped

PR #244, opened 2020-07-13
Isolate docker compose

figassis:
This PR allows running wildduck locally with isolated mongodb and redis resources by providing custom configs


docker-compose.yml
```diff
@@ -10,21 +11,23 @@ services:
...
-            - redis
+            - cache
             - mongo
-
-    redis:
-        network_mode: host
+        volumes:
+            - ./data/dbs.toml:/wildduck/config/dbs.toml
+            - ./data/api.toml:/wildduck/config/api.toml
+    cache:
```
louis-lau, line 19: Not sure if `cache` is the right container name here. Redis stores stuff like sessions. What's wrong with the previous `redis`?

docker-compose.yml
```diff
@@ -10,21 +11,23 @@ services:
...
     mongo:
-        network_mode: host
         image: mongo
+        container_name: mongo
         restart: always
         volumes:
             - /data/db
-        ports:
-            - 27017
+        expose:
```
louis-lau, line 32: The mongo dockerfile already exposes port 27017. I agree with you that it shouldn't be published by default, so I think it can just be removed.

compose.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
louis-lau, line 1: This script feels a bit unnecessary to me. Anyone deploying a container using docker-compose should already know how to do so. I'll let you and andris be the judge of that though.

data/api.toml
```diff
@@ -0,0 +1,56 @@
+enabled=true
```
louis-lau, line 1: Why duplicate some config files? You could mount them directly from the already existing config folder. If any config needs changing it would now need to be changed in multiple places.

andris9:
Same as [this](https://github.com/nodemailer/wildduck/pull/243#issuecomment-657759914)

louis-lau:
I do feel like not publishing the mongodb port (publishing it makes no sense at all and is a security risk when running without auth) and also removing the host networking makes a lot of sense for the main WD repo.

The other changes are subjective.

docker-compose.yml
figassis: If you're working locally, its likely you have redis running for other things, same as mysql and the like. Any other name will do, just to avoid conflicts. I can revert to redis and use wildduck_redis_1 as hostname in config.

docker-compose.yml
figassis: Yes, it can, I had published to check locally, but then hid and just didn't remove to keep things explicit, will remove.

compose.sh
figassis: Force of habit, less typing, will remove

docker-compose.yml
louis-lau: You should always edit a docker-compose to be compatible with your own setup. `redis` is just fine as a name.

data/api.toml
figassis: This is due to disabling host networking, bc the config files are pointing to 127.0.0.1 for redis and mongo. I couldn't find a reasonable way to modify the original configs without making them docker specific with hostnames like mongo and redis.

louis-lau: It may be better to add those specific things as `CMD_ARGS`. That would be a useful addition to the example docker-compose, and you would not need to duplicate the configuration files.

https://docs.wildduck.email/#/in-depth/docker

figassis:
Hopefully this helps, the new configs are dropped and added the data directory to .gitignore, so its generated when running ./compose.sh, but never committed. I can also keep this strictly on my fork :-)

data/api.toml
figassis: ahh, I literally just saw the cmg args in the docs now, totally reinvented the wheel, updating, thanks!

figassis:
Done, much happier now

louis-lau:
I removed the volumes section, as `CMD_ARGS` removes the entire need for it.

Otherwise, looking good👍. Merging.

Issue #142, opened 2019-08-02
SHA265 and SHA512 password hash support


louis-lau:
Heya,

We're looking to migrate from cPanel/WHM to WildDuck with our own (open source) control panel in front. And while I'm amazed you even thought about the ability to add already hashed passwords (very helpful), all of our current passwords are stored as SHA512.

If this is not something you would like to implement I can start messing around an open a PR if needed. But if it _is_ something you'd like to implement, you'll do it a heck of a lot quicker than me. 😁

andris9:
Allowing existing hash formats was added for the exact use case of importing existing users (additional component for importing is the [maildir importer tool](https://github.com/nodemailer/import-maildir)).

If you want to add support for additional hash formats then it should be actually quite easy. All hash validation logic resides [in this file](https://github.com/nodemailer/wildduck/blob/master/lib/hashes.js). If you want to add additional hashes then you would need to update `module.exports.compare` and `module.exports.shouldRehash` methods (`module.exports.hash` should most probably kept as is). 

Password hashes are expected to be formatted like this: `$algo$whatever`.

`'algo'` strings are used in the switch statements. Use the 'md5-crypt' hash handling as an example (algo identifier `'1'` in switch logic).

If you want to have your users hashes converted to default format (on next login), then `shouldRehash` method should return `true` for your algo. If you want to keep the hashes as they are, then return `false`.

If your current stored password hashes are stored as vanilla hashes without any algo prefixes then you should add the prefix to the hash yourself when importing.

louis-lau:
That seems pretty easy indeed. Seems I should be able to implement this. Thanks for the pointers! Want this back as a PR when done?

andris9:
Yes, create a PR for this, might be useful for others as well

PR #245, opened 2020-07-13
add support for sha512crypt hashes during migration

figassis:
This PR allows creating users with hashed passwords with "$6" prefix for the sha512 crypt algorithm. It also marks any such hashes for rehashing back to sha256.


andris9:
Thanks!

louis-lau:
Nice! Closes #142

PR #247, opened 2020-07-14
Sha512 support

figassis:
Improve user creation performance by removing calls to crypto libraries


andris9:
Thanks, seems like a reasonable optimisation

PR #262, opened 2020-08-29
Fix custom 2fa api response

louis-lau:
Currently /users/:user/2fa/custom responds with the following :
```json
{
    "success": {
        "success": true,
        "disabled2fa": true
    }
}
```

This fixes the response to be in line with the docs and all other api methods.


andris9:
Thanks!

PR #266, opened 2020-10-05
Fix disable custom 2FA response

louis-lau:
Similar to #262, wildduck would return this when successfully disabling custom 2FA:
```json
{
  "error": "Failed to enable 2FA",
  "code": "2FAEnableFailed"
}
```

This fixes the issue. Side note, have you ever looked into TypeScript? I personally have come to love it, and it would prevent exactly these kinds of issues.

### DIFF
--- a/lib/api/2fa/custom.js
+++ b/lib/api/2fa/custom.js
@@ -166,19 +166,19 @@ module.exports = (db, server, userHandler) => {
             }
 
             let user = new ObjectID(result.value.user);
-            let { success, disabled2fa } = await userHandler.disableCustom2fa(user, result.value);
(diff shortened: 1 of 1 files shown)


andris9:
TypeScript is great ... until you spend days to overcome some type conflict and everything breaks down until the issue is resolved

louis-lau:
I haven't had that experience 😁. I have felt that with Java before, but I'm sure it's the fact that i'm just not very experienced with it. When exploring wildduck source code I keep finding myself ctrl+clicking to go to definition haha. Javascript makes me feel a bit blinded now. Anyway, it doesn't matter. We all have our reasons for using certain technologies.

PR #267, opened 2020-10-07
Fix api giving ObjectId instead of string

louis-lau:
Unsure what changed to cause this, but the response of `/addresses/resolve/:address` responded with an ObjectId instead of a string for `user` like this:
```json
{
    "success": true,
    "id": "5e7574f5a2bd6304e096a8c2",
    "address": "john@example.com",
    "user": {
        "_bsontype": "ObjectID",
        "id": {
            "type": "Buffer",
            "data": [
                94,
                117,
                116,
                245,
                162,
                189,
                99,
                4,
                224,
                150,
                168,
                188
            ]
        }
    },
    "tags": [],
    "created": "2020-03-21T01:59:17.443Z"
}
```

This fixes it. 😁

### DIFF
--- a/lib/api/addresses.js
+++ b/lib/api/addresses.js
@@ -247,7 +247,7 @@ module.exports = (db, server, userHandler) => {
                         id: addressData._id.toString(),
                         name: addressData.name || false,
                         address: addressData.address,
-                        user: addressData.user,
+                        user: addressData.user.toString(),
                         forwarded: !!addressData.targets,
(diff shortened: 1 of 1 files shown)


louis-lau:
Just noticed the same is true for `/addresses`. Amended my commit and force pushed the PR branch 😁

andris9:
Oh, I thought I handled these. Obviously did not. This is caused by `permission.filter(values)` that filters out properties not allowed by current role. I think it uses JSON internally and that messes up all ObjectId values, so any ObjectId values should be casted to string with toString() before applying the filter.

PR #273, opened 2020-10-18
Filter partial update fix

louis-lau:
Calling `{put} /users/:user/filters/:filter` without either `query` or `action` produced an error similar to these.
```json
{
  "error": "Cannot use 'in' operator to search for 'seen' in undefined"
}
```

Wrapped the existing logic in some checks for `query` and `action`. Also changed `error: 'No changes'` to `success: true`. IMO if you ask for no changes, and then no changes are performed, that's a success.

The github diff doesn't seem to like showing indentation changes very clearly it seems. 😉

### DIFF
--- a/docs/api/api_data.js
+++ b/docs/api/api_data.js
@@ -6893,7 +6893,7 @@ define({ "api": [
           {
             "group": "Parameter",
             "type": "Object",
-            "optional": false,
+            "optional": true,
             "field": "query",
             "description": "<p>Rules that a message must match</p>"
           },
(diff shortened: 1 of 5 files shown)


andris9 (approved):

andris9:
That was one confusing diff. Now that I look at the code I see the mistake. I thought that by applying `.default({})` to the query object in the input processing step would ensure that there is always a query object available but I was not using only the validation result but the original input as well. And the "no changes" error response had a specific reason but I don't remember what the reason was anymore, so I guess there is no reason.

louis-lau:
Ahh, makes sense. Yeah not sure what's up with the diff hahaha, hiding whitespace changes helps. Cheers!

PR #282, opened 2020-11-09
Openapi 3 specification

### LINKED ISSUE #179: OpenAPI/Swagger specification
Having an openapi specification would allow you or anyone else to generate api client libraries automatically: https://openapi-generator.tech/docs/generators.html. I've used this in my code for duckypanel and it's wonderful. You don't have to remember any urls, or response types, as it will generate types as well. You can pass the baseurl and access token once, and not bother with it in the rest of your code.

I've tried converting the apidoc documentation to openapi using [apidoc-swagger](https://www.npmjs.com/package/apidoc-swagger), but it doesn't support openapi 3 and the output contained thousands of errors.

What's your opinion about switching from apidoc to [swagger-jsdoc](https://www.npmjs.com/package/swagger-jsdoc)? You could provide a beautiful documentation page [like this](https://redocly.github.io/redoc/) to replace the generated apidoc as well. Of course this would mean rewriting a lot of stuff, but that could be a PR 😉

louis-lau:
This took quite a lot of hacky scripting and manual work, but I finally have an openapi spec for wildduck!

In #179 I mentioned switching to swagger-jsdoc to have the openapi spec next to each endpoint like the current apidoc. But working with yaml in jsdoc is an absolute nightmare. So I decided to go for a single spec file. It's a bit unwieldy at times, but it does have advantages over apidoc. You can also edit the spec in one of the numerous GUI editors, I like the open source [Stoplight Studio](https://stoplight.io/studio/).

New api docs:
https://louis-lau.github.io/wildduck/api/

Example Node.js api client I generated from the spec:
https://github.com/louis-lau/wildduck-example-api-client

Cool things you can do with an openapi spec:
https://openapi.tools/

Let me know if you see any mistakes! Hard to catch them all with such a massive spec. 😁 

Fixes #179


louis-lau:
Ideally all parameters should have examples, but that's something for the future :)

andris9:
This looks great! The only thing I'm slightly worried about is keeping the definition file in sync with code updates. There are some netlify build errors, what are these about? Could you fix these before merging?

louis-lau:
Yeah that was my concern as well, I really did try to keep it next to the code haha, but it just wasn't practical. 

Looks like netlify is trying to build the apidoc, but it can't since I removed the dependency.
```
5:46:21 PM: $ npm run apidoc
5:46:21 PM: > wildduck@1.30.3 apidoc /opt/build/repo
5:46:21 PM: > apidoc -i lib/api/ -o docs/api/
5:46:21 PM: sh: 1: apidoc: not found
```

I don't think the netlify config is in the repo, is it?

louis-lau:
While I'm used to [Nests openapi Integration](https://docs.nestjs.com/openapi/introduction) for generating it, I've read that other teams update the openapi doc _before_ making any code changes. Could be a good strategy :grin:

andris9:
Now that I think of it, Netlify support isn't even needed. I think it was just an experiment to publish docs there instead of Github. I removed Netlify from integrations.

louis-lau:
Ha, I was a little confused as well as I thought the docs used github pages. Glad to see this merged 😁.

PR #292, opened 2020-11-18
Add example values for getUserAddressregister

louis-lau:
Ideally should be done for all endpoints, but it's easier to start with the newly added ones 😁 

Makes the examples in the doc a lot more rich:
image

Yes, I'm a bit bored today :)


andris9:
Thanks. This specific feature might change still. It's for the type ahead feature of webmail, so when you would start typing recipient address then it would load addresses from this endpoint and "suggest" these for auto completion.

PR #330, opened 2021-04-04
Various openapi doc fixes

louis-lau:
This PR fixes various issues with the openapi doc


louis-lau:
@andris9 This is ready to merge if you agree with all these fixes

PR #332, opened 2021-04-05
Api: Fix cors config being required to start

### LINKED ISSUE #331: wildduck container goes into restart loop directly after setup
```
/wildduck/api.js:147

origins: [].concat(config.api.cors.origins || ['*']),
                                   ^

TypeError: Cannot read property 'origins' of undefined
```

networkException:
This makes the cors config table entirely optional, instead
of previously just the origins key inside the cors table

Resolves #331


networkException:
Would you at least have the decency to not mark prs as stale that are lacking attention from the maintainers?

Issue #33, opened 2017-10-29
Pluggable storage ?


robbiehanson:
I understand that wildduck is "opinionated". However, is there any chance you'll be supporting pluggable storage in the future?

I notice that other projects (based on your work) have sprung up specifically to address this concern: [imapper](https://github.com/deitch/imapper)

We're keen to use wildduck. But pluggable authentication and pluggable storage are requirements I don't think we'd be able to work around.

andris9:
Pluggable authentication is doable. I'm experimenting with a plugin system for Wild Duck and authentication is just a call that takes an username and a password as an input and should yield with an user ID. It does get more complicated with the SMTP plugin as it performs authentication directly against the database but I guess it wouldn't be difficult to edit it in a way to perform auth against a Wild Duck server and not against the database.

Pluggable storage is another story. It would be possible to use a custom storage for attachments (Wild Duck separates attachments from the message), as all it requires is an object store. In fact the current [attachment store](https://github.com/nodemailer/wildduck/blob/master/lib/attachments/gridstore-storage.js) is already organised as kind of a plugin as it is decoupled from the rest of the code and could just as well store the attachments to S3 for example.

Pluggable message store is much more difficult to do though. The number one goal for Wild Duck is scalability, so I have spent a lot of time evaluating different storage options and have done a lot of MongoDB specific tuning. There is nothing in there too MongoDB specific though that could not be replaced with something else even if the replacement would be a huge task but as a sole developer my time resources are extremely limited and if something does not get me closer to my own goals, then I can not work on it.

If it's not the actual store that you care about but want to build something very specific, for example an encrypted message store or dynamic message store where the IMAP mailbox would be a front end to something else, then I guess that the best bet would be to take the code from [imap-core folder](https://github.com/nodemailer/wildduck/tree/master/imap-core) and build in top of that. [Here's a minimal IMAP server](https://github.com/nodemailer/wildduck/blob/master/imap-core/test/test-server.js) built on top of the core. You could take that example server and build your custom logic on top of it.

andris9:
Just for the reference, to get the test server running:

```
$ git clone git://github.com/nodemailer/wildduck.git
$ cd wildduck
$ npm install --production
$ node -e 'require("./imap-core/test/test-server.js")({secure:true}).listen(9993)'
```

and then from another tab

```
$ openssl s_client -crlf -connect localhost:9993
> * OK wildduck ready
```

You can then use usernamer 'testuser' and password 'pass' to log in:

[...]

PR #337, opened 2021-04-26
added base changes to support s3

bhanuc:
Aim: to add support for s3 and other s3 compatible storage layers(Like DO Spaces etc) . Issue was already discussed in #33 . 

This PR is aimed at getting primary feedback on the design and feedback and also gather requirements/needs this project might have.

Also I have not tested it fully and I plan to write tests once I get some feedback on this.

### DIFF
--- a/api.js
+++ b/api.js
@@ -397,7 +397,6 @@ module.exports = done => {
     if (!config.api.enabled) {
         return setImmediate(() => done(null, false));
     }
-
     let started = false;
 
     const component = config.log.gelf.component || 'wildduck';
@@ -447,11 +446,12 @@ module.exports = done => {
         attachments: config.attachments,
         loggelf: message => loggelf(message)
     });
-
+    
     storageHandler = new StorageHandler({
         database: db.database,
         users: db.users,
         gridfs: db.gridfs,
+        config: config.storage,
         loggelf: message => loggelf(message)
     });
 
--- a/config/default.toml
+++ b/config/default.toml
@@ -83,6 +83,9 @@ maxForwards=2000
 [plugins]
 # @include "plugins/*.toml"
 
+[storage]
+# @include "storage.toml"
+
 [tasks]
 # if enabled then process jobs like deleting expired messages etc
 enabled=true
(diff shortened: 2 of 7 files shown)


andris9:
Well yes, the storage system was initially meant to support different providers like that. 

Adding all these storage drivers into WildDuck though would end up with a huge amount of unneeded dependencies (eg. someone using standard GridFS still has to install AWS SDK) so a better approach would treat these mechanisms as plugins.

So what would need to be different:

* GridFSDriver as a standard would be baked in
* The storage handler driver would either require the built in gridfs driver or an external dependency based on cofngiguration (eg `const SelectedDriver = require('wildduck-s3-driver')`)

Plugin's package.json would list WildDuck as a peer dependency

```json
{
  "name": "wildduck-s3-driver",
  "peerDependencies": {
    "wildduck": "1.33.2"
  }
}
```

So basically everything would be the same way you already did this except that the S3 driver would be a separate plugin.

andris9:
(using fixed peer dependency versions number is not a good idea though.  Having something like "1.x" would probaby be better. [Here's](https://nodejs.org/es/blog/npm/peer-dependencies/) some more information)

andris9:
Btw this PR only covers Draft composing storage but not actual attachments. The main differences between attachments and draft storage are:

* attachments are deduplicated, draft storage is not. This is not a hard requirement though as the deduplication is handled by the driver. GridFS does this by using sha256 hash of the attachment file as the ID. You could just as well generate a new ID for each attachment and ignore deduplication entirely.
* attachments are stored as decoded binary files. Yet again, not a hard requirement but allows to keep storage size down (base64 encoded attachment takes at least 33% more storage than a decoded binary file)
* attachment content can be accessed partially. This is need by IMAP where clients tend to do partial message downloads. Clients sometime ask messages in small chunks so you do not want to fetch a 20MB attachment just to return the requested 64kB part of it. Requested range is for the original form of the attachment, so if you have decoded attachment into a binary file then you have to calculate actual bounds to get the needed part for base64 content which is tricky but the current GridFS attachment storage handles this alreay.

PR #347, opened 2021-08-06
Add idate to EXISTS update entry

louis-lau:
When a new message comes into the inbox, I want to decide whether it should be downloaded or not based on what messages are currently in the view, and what messages are yet to be fetched because they're on the next page.

Since the message list is sorted by the idate, adding it to the exists event should make this possible for me :)


andris9:
Seems reasonable

PR #349, opened 2021-08-30
DKIM: changed encoding of public and private keys

kubuntux:
I am having a hard time making DKIM generated by the API to work in gmail.

Screen Shot 2021-08-30 at 1 31 06 PM

I found this site to verify DKIM public key: https://dkimcore.org/c/keycheck

All keys being generated by the API appears to be invalid based on the site above.

Screen Shot 2021-08-30 at 1 34 33 PM

I made changes in the encoding when generating, which is also what is noted in this page: https://stackoverflow.com/questions/51942824/nodejs-generate-valid-pem-keys-for-signing-and-verifying-messages

Now it worked:

Screen Shot 2021-08-30 at 1 30 48 PM

Screen Shot 2021-08-30 at 1 33 59 PM

### DIFF
--- a/lib/dkim-handler.js
+++ b/lib/dkim-handler.js
@@ -30,11 +30,11 @@ class DkimHandler {
             modulusLength: keyBits || 2048, // options
             publicExponent: keyExponent || 65537,
             publicKeyEncoding: {
-                type: 'pkcs1',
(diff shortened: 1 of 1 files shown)


andris9:
Thanks!

PR #353, opened 2021-09-12
Add optional thread message counters to listing

louis-lau:
I want to have a list view with threading, but I don't want to fetch the messages for a specific thread if there's only a single message in the thread. Overhead shouldn't be too great as I see it, but I made it optional anyway to avoid it.

### DIFF
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -1503,6 +1503,11 @@ paths:
                   description: 'If true, then includes metaData in the response'
                   schema:
                       type: boolean
+                - name: threadCounters
+                  in: query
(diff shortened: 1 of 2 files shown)


louis-lau:
Well this is stupid. I guess I was tired when coding this. But thread.ids.length does not correspond to the amount of messages in a thread. Do you have any suggestions on how to count this efficiently @andris9?

louis-lau:
Now works correctly. It just counts how many messages there are for eacht thread using an aggregation. Uses existing indexes so it's pretty fast :)

andris9:
Llooks good

PR #360, opened 2021-11-13
Fixed bug where ssl certs without a secret are not working probably

Justin8303:
SSL certificates that are not encrypted with a secret are not working.
Instead, an error occurs because the provided secret is undefined.
This pull request fixes the bug when a "cleartext" SSL certificate is provided.


andris9:
Thanks!

PR #392, opened 2022-04-20
Fixed Webauthn registration failing.

jissepo:
The webauthn registration logic expects the rpId / rp.id value to be a valid eTLD+1. Because currently the user passed origin doesn't get passed into Fido2Lib the returned challenge object doesn't contain a valid domain to register the credentials for. 

This could be bypassed by removing the rp.id key from the returned object, but the attestation fails later with the error "origin is not a valid eTLD+1". This PR should fix both issues.

Maybe we should also add the abilty to pass in rpName value, in case in the future a single wildduck instance gets used for multiple email clients.

Also maybe should rename origin now that it no longer actually is origin.


jissepo:
This is incorrect, the eTLD+1 error comes from using .test domain as local dev environment domain

PR #393, opened 2022-05-05
Sort message search by date

louis-lau:
Unless there's a good reason to sort these by id?


andris9:
The only concern is indexing. `_id` gives an ordered index for "free". Maybe you could run some `explain`s and verify that MongoDB always uses a proper index for these updated queries.

louis-lau:
Sure! From my quick testing queries didn't take any longer than before. Butt I'll check to be sure.

andris9:
And maybe also do it like this: if ordering is requested, then use the date field, if it was not requested, then use the _id as it's right now.

louis-lau:
@andris9 Sorry, kinda forgot about this one. I used mongodb profiling to figure out the queries that `mongo-cursor-pagination` is running, ran explain on them, and the only difference is the memory usage. The current implementation does not use the index for sorting either.

memUsage of SORT stage for 250 results on my test account:
current: 18MB
this PR: 23MB

It's now implemented in a way where it only sorts by date if you supply an order.

andris9:
Great, thanks!

PR #433, opened 2022-11-07
Fix 05_install_packages.sh for python3

RealCalumPlays:
This should fix the script installing python as python isnt a valid package on ubuntu 22.04 and im pretty sure its not valid on 20.04 now.


RealCalumPlays:
Closed as I found other issues

PR #437, opened 2022-11-17
Issue 436: POP3 AUTH PLAIN not working for certain username and password lengths

KZumbusch:
Replacing the regular expression should fix the issue with POP3 logins with certain password lengths


andris9:
Thanks!

I've tried to use `Joi` library to validate all formats, so I wonder why I went with that strange regex instead of using `Joi.string().base64()` 🤔

Issue #432, opened 2022-11-04
Username with @ fails to resolve (InputValidationError)


zdeneksvarc:
Looks like a sanitization error.

```
> curl -H 'Content-Type: application/json' -H 'X-Access-Token: ...' https://api/users/resolve/hello%40acme.com

{
  "error": "\"username\" with value \"hello@acme.com\" fails to match the username pattern",
  "code": "InputValidationError",
  "details": {
    "username": "\"username\" with value \"hello@acme.com\" fails to match the username pattern"
  }
}
```

The same without URL encoding.

louis-lau:
Perhaps you're looking for the `/addresses/resolve/{address}` endpoint? Usernames are only alphanumeric + dots.

zdeneksvarc:
Yes, now I see `Username of the User. Alphanumeric value. Must start with a letter, dots are allowed but informational only ("user.name" is the same as "username")`

I'm trying to have the username as an address, including the @. I can create such a username, I can list it among users, but I can't resolve the ID.

louis-lau:
I see. I would have assumed that you wouldn't be able to set that username, but if you can then this validation is wrong indeed :)

I do wonder about your usecase though, resolving the address in this case seems much more useful right?

zdeneksvarc:
Thank you for your interest. I am testing API endpoints in preparation for a simple Deno CLI application that would allow easy user management.

Since I'm used to address-shaped usernames, I didn't actually realize that Wild Duck is different in this regard. It's my fault for inattention.

louis-lau:
The idea was taken from gmail, where if your address is `john@example.com`, you can also log in with the username `john`. The nice thing is that you can also change the address associated with the user entirely, but keep the username the same.

zdeneksvarc:
Yes, that's exactly right. It's just a mind-set, nothing fundamental.

zdeneksvarc:
The behavior is even stranger. If I create a user with @, which the API allows, then I am unable to resolve the id. Again, for the reasons mentioned above (InputValidationError).

Well, if I'm unable to resolve the id, then I can neither update nor delete the user because these API calls need a user id. Of course, I can list all the users and grep. But that's dirty unsystematic hack. Any thoughts please?

louis-lau:
As I said above:
> this validation is wrong indeed

So it would need to be fixed :)

Though if you're writing a users management app, I'd have thought you're already listing the users and already should have the ids. But I'm not sure how your app works of course.

[...]

Issue #414, opened 2022-08-02
Cannot resolve users with "@" in the username


ramiroaisen:
I have configured usernames in my setup of wildduck to be the same as their corresponding addresses.

I cant use the `/users/resolve/{username}` endpoint as it fails with the following error.

`GET /users/resolve/address@example.com`
```json
{
  "error": "'username' with value 'address@example.com' fails to match the username pattern",
  "code": "InputValidationError",
  "details": {
    "username": "'username' with value 'address@example.com' fails to match the username pattern"
  }
}
```

Wildduck version: 1.35.6

PR #454, opened 2023-03-27
Update users.js

mehmet-kozan:
resolve username which contains "@", regex same as "post:user/" endpoint. related #414 #432


andris9:
Thanks!

PR #455, opened 2023-03-28
Update users.js

mehmet-kozan:
user:resolve if username setted as an email. this code removes domain dots also.  fixed via using tools.normalizeAddress function. related related #414 #432 #454

### DIFF
--- a/lib/api/users.js
+++ b/lib/api/users.js
@@ -585,7 +585,10 @@ module.exports = (db, server, userHandler, settingsHandler) => {
             try {
                 userData = await db.users.collection('users').findOne(
                     {
-                        unameview: username.replace(/\./g, '')
+                        unameview: tools.normalizeAddress(username, false, {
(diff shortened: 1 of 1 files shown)


mehmet-kozan:
`tools.normalizeAddress(username, false, {
                        removeLabel: true,
                        removeDots: true
                    })`
**username input <-> output must be**
"mehmet.kozan@abc.com" >  "mehmetkozan@abc.com"
"mehmet.kozan" > "mehmetkozan"
"mehmet" > "mehmet"

**actual results**
"mehmet.kozan@abc.com" >  "mehmetkozan@abc.com"
"mehmet.kozan" > "@mehmet.kozan"
"mehmet" > "@mehmet"


if tools.normalizeAddress function works correctly on non-email usernames, the if-else block none needed

andris9:
`normalizeAddress` is for email addresses, not usernames, so it works correctly.

PR #496, opened 2023-09-14
Grunt mocha test fixes

NickOvt:
Instead of having just one subtask (subobject) on the mochaTest command that does just one test run - separate the imap-core and api test runs into separate subtasks. That means both test runs will be separate from each other


andris9 (approved):
LGTM

PR #503, opened 2023-09-22
Zms 90

NickOvt:
1. Create, Update, Delete functions in mailboxes-handler.js updated to have better and more informative error output
2. Tests updated accordingly
3. mailboxes.js API file also updated and unnecessary parts are removed

### DIFF
--- a/lib/mailbox-handler.js
+++ b/lib/mailbox-handler.js
@@ -125,10 +125,16 @@ class MailboxHandler {
                     return callback(err);
                 }
                 if (!mailboxData) {
-                    return callback(null, 'NONEXISTENT');
+                    const err = new Error('Mailbox update failed with code NoSuchMailbox');
+                    err.code = 'NoSuchMailbox';
+                    err.httpStatus = 404;
+                    return callback(err, 'NoSuchMailbox');
                 }
                 if (mailboxData.path === 'INBOX' || mailboxData.hidden) {
-                    return callback(null, 'CANNOT');
+                    const err = new Error('Mailbox update failed with code DisallowedMailboxMethod');
+                    err.code = 'DisallowedMailboxMethod';
+                    err.httpStatus = 500;
+                    return callback(err, 'DisallowedMailboxMethod');
                 }
                 this.database.collection('mailboxes').findOne(
                     {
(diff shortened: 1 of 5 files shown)


NickOvt:
IMAP Protocol integration tests
       CREATE
         should create new mailbox:
         
        Uncaught AssertionError: expected false to be true
        + expected - actual

        -false
        +true

This test seems to fail for some reason now, will investigate.
My idea is that probably because I changed the status codes (the wording) this test fails.

andris9:
> ```
> IMAP Protocol integration tests
>        CREATE
>          should create new mailbox:
>          
>         Uncaught AssertionError: expected false to be true
>         + expected - actual
> 
>         -false
>         +true
> ```
> 
> This test seems to fail for some reason now, will investigate. My idea is that probably because I changed the status codes (the wording) in the mailboxes-handler.js this test fails.

Uppercase error codes are special IMAP responses, these must stay the same. Not sure what to do with these, maybe create some kind of wrapper than normalized IMAP error codes for more readable codes for API responses 🤔

NickOvt:
in file `imap-core/lib/commands/create.js` the function `create` always returns `TEMPFAIL` instead of the fail returned by the server, thus the given test fails. Any ideas on fix? If TEMPFAIL error response needs to stay then we can change the test appropriately OR make it that create returns the `err.code`

NickOvt:
Also fixed imap-core create function so it returns correct error instead of hardcoded TEMPFAIL value

lib/mailbox-handler.js
```diff
@@ -201,10 +222,16 @@ class MailboxHandler {
...
+                    const err = new Error('Mailbox deletion failed with code NoSuchMailbox');
+                    err.code = 'NONEXISTENT';
+                    err.responseCode = 404;
+                    return callback(err, 'NONEXISTENT');
                 }
                 if (mailboxData.specialUse || mailboxData.path === 'INBOX' || mailboxData.hidden) {
-                    return callback(null, 'CANNOT');
+                    const err = new Error('Mailbox deletion failed with code DisallowedMailboxMethod');
+                    err.code = 'CANNOT';
+                    err.responseCode = 405;
```
andris9, line 233: 405 is not a good choice, even if the status response message would seem like it. 405 covers HTTP methods, not API methods, i.e. you should get it when running PUT to a path that only allows GET requests. There are also [additional requirements](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405) for using this response code (has to return a `Allow` header)

Use a generic error code like 400 instead for the 405 responses.

andris9 (changes requested):

lib/mailbox-handler.js
NickOvt: Fixed. Made 405 errors to be 400 errors, fixed tests accordingly.

andris9 (approved):

PR #535, opened 2023-10-16
ZMS-[9x] Automatic API generation

NickOvt:
- [x] ZMS-96 - Add additional data to Endpoints in the message.js file (Messages) to include them in API generation
- [x] ZMS-97 - given an object that has key-value pairs where the value is a Joi validation object, parse that object and build a simple JS object from it with required data to generate a YAML string required for automatic API generation
- [x] ZMS-98 - Actual API generation  
Every endpoint should be (but doesn't have to, in which case it will be ignored) converted to a new format, that will allow the given endpoint to be used for automatic API generation if the endpoint is allowed to be included in API generation,
Data for every endpoint:
```js
{
  path: "api route path",
  pathParams: {
    [pathParamKey]: <simple Joi object (primitive data type)>,
  },
  summary: "summary",
  description: "description",
  requestBody: {
    [requestBodyParamKey]: <Joi object>,
  },
  applicationType: "application/json or any other valid type", // if not given defaults to "application/json"
  queryParams: {
    [queryParamKey]: <simple Joi object (primitive data type)>
  },
  tags: ["Tag1"],
  response: {
    [httpStatusCode]: {
      description: "description",
      model: <Joi object>
    }
  },
  responseType: "application/json or any other valid type", // if not given defaults to "application/json"
  include: boolean // include or exclude the API Route/Endpoint from API generation. If not given then regarded as false
}
```
The object above ^ has to be the first parameter in the `server.get, server.post, server.put, server.del` function calls.

1. Define constant parts of the documentation. Such as tags, servers, info, auth, etc.
2. Go through all API Routes
3. Extract necessary data from the endpoint data, such as `tags, summary, description, operationId, requestBody, parameters, responses`
4. Parse all Joi objects that come along the way, if necessary parse Joi objects recursively if the object contains inner objects or it is an array or an `alternatives` Joi type
5. Given all data above start building the initial documentation in JS object format.
6. After the documentation initial JS object is built walk through it (if necessary recursively) to extract repeating components into separate components and replace the part with a ref to the component. (Used to comply with modern openApi documentation standards and best practices)
7. Transform JS object to YAML string with help of `js-yaml` package. Also as components are held in a separate JS object transform that object to YAML string as well.
8. Concatenate all YAML strings into one string
9. Write docs YAML string to a file (currently it goes to the root of the project and the file name is `openapidocs.yml`)


lib/tools.js
```diff
@@ -16,9 +16,15 @@ const ipaddr = require('ipaddr.js');
 const ObjectId = require('mongodb').ObjectId;
 const log = require('npmlog');
 const addressparser = require('nodemailer/lib/addressparser');
+const yaml = require('js-yaml');
 
 let templates = false;
 
+if (typeof structuredClone !== 'function') {
+    // eslint-disable-next-line no-global-assign
+    structuredClone = obj => JSON.parse(JSON.stringify(obj));
```
andris9, line 25: All variables must be defined, so this line throws an error:

```
ReferenceError: structuredClone is not defined
    at Object.<anonymous> (/home/runner/work/wildduck/wildduck/lib/tools.js:25:21)
    at Module._compile (node:internal/modules/cjs/loader:1198:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1252:10)
    at Module.load (node:internal/modules/cjs/loader:1076:32)
    at Function.Module._load (node:internal/modules/cjs/loader:911:12)
    at Module.require (node:internal/modules/cjs/loader:1100:19)
    at require (node:internal/modules/cjs/helpers:119:18)
    at Object.<anonymous> (/home/runner/work/wildduck/wildduck/lib/imap-notifier.js:4:15)
    at Module._compile (node:internal/modules/cjs/loader:1198:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1252:10)
```
It is probably better to create a local variable under a different name that either gets the existing `structuredClone` (if it is set) or falls back to the JSON handler

```js
const structuredCloneWrapper = typeof structuredClone === 'function' ? structuredClone : JSON.parse(JSON.stringify(obj));
```

andris9 (changes requested):

lib/tools.js
```diff
@@ -694,5 +870,202 @@ module.exports = {
                 res.json(data);
             }
         };
+    },
+
+    async generateAPiDocs(routes) {
+        let docs = `
+openapi: 3.0.0
```
louis-lau, line 877: openapi spec doesn't have to be yaml, it can be JSON. I just used yaml because it was easier to edit by hand. I think it would make more sense here to just work with js objects/arrays normally and output a json file at the end right? Seems like it would be much more rebust than appending to a yaml string.

lib/tools.js
```diff
@@ -694,5 +870,202 @@ module.exports = {
...
+            if (spec.exclude) {
+                continue;
+            }
+
+            if (!mapPathToMethods[spec.path]) {
+                mapPathToMethods[spec.path] = {};
+            }
+
+            mapPathToMethods[spec.path][spec.method.toLowerCase()] = {};
+            const methodObj = mapPathToMethods[spec.path][spec.method.toLowerCase()];
```
louis-lau, line 929: Since you're constructing an openapi operation here (I think?), is methodObj really a clear variable name? I think it would make sense to follow the same naming as the openapi spec in all of this for code readability.

lib/tools.js
```diff
@@ -694,5 +870,202 @@ module.exports = {
...
+            // 1) add tags
+            methodObj.tags = spec.tags;
+
+            // 2) add summary
+            methodObj.summary = spec.summary || '';
```
louis-lau, line 934: Almost none of the fields here that you use `''` as a fallback for in this whole function are actually required by the openapi spec. The fallback doesn't really make sense in any of these cases, it would just add clutter to the final spec.

louis-lau (commented):
Andris is in charge of all application logic, but I though my feedback could help here 😁 

Should this maybe PR be marked as draft until it's actually ready?

lib/tools.js
NickOvt: Yes. You are correct. In the upcoming commits I am planning to make it all JSON, that way it's more standardized and more easily maintainable! Also more robust as you have pointed out

lib/tools.js
NickOvt: Honestly, added the fallback so there's less if statements and made an assumption that the clutter in the final spec is unimportant as it's automatically generated anyway. But it seems to be a good idea to remove the fallbacks indeed, that way the final doc will become more readable.

lib/api/messages.js
```diff
@@ -1983,18 +1961,52 @@ module.exports = (db, server, messageHandler, userHandler, storageHandler, setti
...
+                            success: Joi.boolean(),
+                            queueId: Joi.string(),
+                            forwarded: Joi.array().items(
+                                Joi.object({
+                                    seq: Joi.string(),
+                                    type: Joi.string(),
+                                    value: Joi.string()
+                                }).$_setFlag('objectName', 'Forwarded')
+                            )
+                        })
```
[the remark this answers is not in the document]
NickOvt: There are a couple things I am still thinking of adding but highly likely not with this given PR:
1. Joi has checks for hex, binary as well as string max and min length and all of this seems to be supported by OpenAPI as well so adding that would add even more clarity to the inputs (as well as outputs)
2. Noticed that some string inputs in the current doc are actually string `Enum`s, so I am planning to add that as well, currently that functionality is missing and even if string is an enum it will be displayed as generic string type in the docs

Support for example strings would be nice indeed, but currently am not really aware of whether Joi has that kind of thing built in, perhaps will just have to add it do description and hope that the renderer renders it correctly. Will research!

Other responses besides status 200 are to be supported later on as well actually, but they require more work (or maybe not :D, not sure yet about how to best structure them)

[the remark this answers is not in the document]
NickOvt: > Additionally, being able to specify a custom openapi schema would be nice. It wouldn't be used normally, but there's some edgecases where you need the openapi spec to reflect something that's just not possible to define with Joi.

Could be added, sure

lib/tools.js
louis-lau: That's true! But you never know how tools will handle the empty strings hehe. If statements wouldn't be necessary here I'd think right?

lib/api/messages.js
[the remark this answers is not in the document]
louis-lau: Yeah, enums are the way you can define "it's one of these simple values" in an openapi spec. It's very helpful for docs and typed languages!

I did find this:
https://joi.dev/api/?v=17.9.1#anyexampleexample-options

The doc tool we use will use these openapi examples on the right side, instead of something like `email: 'string'` it'd automatically fill it with `email: 'john@example.com'`. It works quite nicely!

NickOvt:
ZMS-96 added all POST and GET endpoints in messages.js to API generation, but as the process of adding is extremely tedious and time consuming, it is better to import existing data from the current API docs and later on just add/remove data.

louis-lau:
For conversion from the old docs to openapi I also used an automatic process, with a lot of manual fixups afterwards :)

andris9 (approved):

PR #542, opened 2023-10-20
ZMS-93

NickOvt:
- `mailbox-handler.js` added `createAsync` as the rewritten and modern version of original `create` function.
support for legacy callback `create` is still there so no older code breaks.
- Now there is a hard limit on mailboxes count for a single user: `1500` mailboxes
- Added new error message (both IMAP error and API error) respectively `"OVERQUOTA"` and `"ReachedMailboxesCountLimit"`

### DIFF
--- a/lib/mailbox-handler.js
+++ b/lib/mailbox-handler.js
@@ -3,6 +3,7 @@
 const ObjectId = require('mongodb').ObjectId;
 const ImapNotifier = require('./imap-notifier');
 const { publish, MAILBOX_CREATED, MAILBOX_RENAMED, MAILBOX_DELETED } = require('./events');
+const { SettingsHandler } = require('./settings-handler');
 
 class MailboxHandler {
     constructor(options) {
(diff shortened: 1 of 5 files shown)


lib/mailbox-handler.js
```diff
@@ -19,99 +20,92 @@ class MailboxHandler {
...
-                            if (err) {
-                                if (err.code === 11000) {
-                                    const err = new Error('Mailbox creation failed with code MailboxAlreadyExists');
-                                    err.code = 'ALREADYEXISTS';
-                                    err.responseCode = 400;
-                                    return callback(err, 'ALREADYEXISTS');
-                                }
-                                return callback(err);
-                            }
+        const userData = await this.database.collection('users').findOne({ _id: user }, { projection: { retention: true } });
```
andris9, line 52: This should probably be the first check - if the user does not exist, there is no need to perform other operations.

lib/mailbox-handler.js
```diff
@@ -19,99 +20,92 @@ class MailboxHandler {
...
-                    return callback(err);
-                }
-                if (mailboxData) {
-                    const err = new Error('Mailbox creation failed with code MailboxAlreadyExists');
-                    err.code = 'ALREADYEXISTS';
-                    err.responseCode = 400;
-                    return callback(err, 'ALREADYEXISTS');
-                }
+        this.createAsync(user, path, opts)
+            .then(mailboxData => callback(null, ...Object.values(mailboxData)))
```
andris9, line 29: The ordering for object keys is not always obvious, and it is better to use specific values and not depend on the ordering of keys in the method response. This seems something that might break in the future where seemingly a simple change (like adding an additional property or using Object.assign) will mess up the current key order.

Just for the reference, this is how object keys are enumerated:
> Within each component of the prototype chain, all non-negative integer keys (those that can be array indices) will be traversed first in ascending order by value, then other string keys in ascending chronological order of property creation.

lib/mailbox-handler.js
```diff
@@ -19,99 +20,92 @@ class MailboxHandler {
...
-                            retention: userData.retention
-                        };
+        const mailboxCountForUser = await this.database.collection('mailboxes').countDocuments({ user });
 
-                        Object.keys(opts || {}).forEach(key => {
-                            if (!['_id', 'user', 'path'].includes(key)) {
-                                mailboxData[key] = opts[key];
-                            }
-                        });
+        if (mailboxCountForUser > (await this.settingsHandler.get('const:max:mailboxes', {}))) {
```
andris9, line 45: The second argument (options object) is optional so that the empty object can be skipped. I have no idea why it is used in other places. Maybe there were some options set initially and later got removed, but the empty options object remained in place 🤔

lib/mailbox-handler.js
```diff
@@ -19,99 +20,92 @@ class MailboxHandler {
...
+            const err = new Error('Mailbox creation failed with code ReachedMailboxCountLimit');
+            err.code = 'OVERQUOTA';
```
andris9, line 47: `OVERQUOTA` has a specific meaning (see [rfc9208#4.3.1](https://www.rfc-editor.org/rfc/rfc9208.html#name-overquota)) so a more generic error like `CANNOT` would be better because this specific error case is not defined (see CANNOT definition in [rfc5530](https://www.rfc-editor.org/rfc/rfc5530#page-5)).

lib/tools.js
```diff
@@ -661,6 +661,10 @@ module.exports = {
                         err.responseCode = err.responseCode || 404;
                         err.code = 'NoSuchMailbox';
                         break;
+                    case 'OVERQUOTA':
```
andris9, line 664: See the comment in https://github.com/nodemailer/wildduck/pull/542/files#diff-786172e327ab72ee48fd40a2bad6f49f157611bd64d1b253b53f596688ad656bR47

`OVERQUOTA` should not return an error code for mailbox count, but, if used, a disk-size related error.

andris9 (changes requested):
Check the comments. In general, seems great 👍

lib/mailbox-handler.js
NickOvt: Made it the first check

lib/mailbox-handler.js
NickOvt: Removed unnecessary second parram. Indeed it is not required and everything seems to work just fine without it

lib/mailbox-handler.js
NickOvt: Fixed

lib/tools.js
NickOvt: Removed from tools.js

lib/mailbox-handler.js
```diff
@@ -26,7 +26,7 @@ class MailboxHandler {
 
     create(user, path, opts, callback) {
         this.createAsync(user, path, opts)
-            .then(mailboxData => callback(null, ...Object.values(mailboxData)))
+            .then(mailboxData => callback(null, ...[mailboxData.status, mailboxData.id]))
```
andris9, line 29: I see that you prefer to use the spread operator, even if you don't really need it 😄

andris9 (approved):

lib/mailbox-handler.js
NickOvt: Ah yes haha. Indeed here it is not required, I could just pass the parameters as they are. Will fix in a sec

andris9: No need, already merged

PR #545, opened 2023-10-26
Zms 91

NickOvt:
API message submission now logs to graylog the added new data. `_limit_sent` and `_limit_allowed`.
Also `to` field is now singular for every `to` address. So for each message there is a **separate** log.
Changed short_message to: `[RCPT TO: ${to}] ${session}`.
Mail action is now `rcpt_to`


andris9 (approved):
LGTM

PR #551, opened 2023-11-08
ZMS-99

NickOvt:
1. Do not add noreply@ and no-reply@ emails to the address register
2. emails containing list-id, auto-submitted, x-auto-response-suppress headers (one of the provided) are skipped for the address register additions completely.

### DIFF
--- a/lib/message-handler.js
+++ b/lib/message-handler.js
@@ -538,7 +538,7 @@ class MessageHandler {
                                                     for (let key of keyList) {
                                                         if (parsed[key] && parsed[key].length) {
                                                             for (let addr of parsed[key]) {
-                                                                if (addr.address.includes('noreply') || addr.address.includes('no-reply')) {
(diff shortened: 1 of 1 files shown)


lib/message-handler.js
```diff
@@ -525,9 +526,21 @@ class MessageHandler {
...
+                                                        // don't add any emails to address register
+                                                        if (parsed[disallowedHeader]) {
+                                                            return next();
+                                                        }
+                                                    }
+
                                                     for (let key of keyList) {
                                                         if (parsed[key] && parsed[key].length) {
                                                             for (let addr of parsed[key]) {
+                                                                if (addr.address.includes('noreply') || addr.address.includes('no-reply')) {
```
andris9, line 541: Maybe use case insensitive regex here?
```
/no-?reply/i.test(addr.address)
```

andris9 (changes requested):

andris9 (approved):

PR #557, opened 2023-11-22
fix: fixed bug with CHECK command not including Authenticated state (closes #556)

### LINKED ISSUE #556: Bug: `CHECK` command not working properly (should not error if authenticated)
Connection state should be either "Authenticated" or "Selected".

Right now if you try to use `himalaya` and run `himalaya account sync` you will get an error with any WildDuck server (it will not sync _any_ messages) due to this.

Submitting a PR shortly.

titanism:
See <https://github.com/nodemailer/wildduck/issues/556> for insight.

### DIFF
--- a/imap-core/lib/commands/check.js
+++ b/imap-core/lib/commands/check.js
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = {
-    state: ['Selected'],
+    state: ['Authenticated', 'Selected'],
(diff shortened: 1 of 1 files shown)


andris9:
`CHECK` is only allowed in Selected state, see [rfc3501#section-6.4](https://datatracker.ietf.org/doc/html/rfc3501#section-6.4)

PR #559, opened 2023-11-22
fix: allow CLOSE command to return gracefully (closes #558)

### LINKED ISSUE #558: Bug: `CLOSE` not allowed now (shouldn't it just return normally even if not selected?)
Similar to https://github.com/nodemailer/wildduck/issues/556, when using `himalaya account sync` it invokes `CLOSE` command when Authenticated (and no mailbox selected).  Instead of throwing an error, shouldn't we just instead return OK if no mailbox selected?

For example:

```diff
module.exports = {
-    state: 'Selected',
+    state: ['Authenticated', 'Selected' ],

    handler(command, callback) {
        // Check if EXPUNGE method is set
        if (typeof this._server.onExpunge !== 'function') {
            return callback(null, {
                response: 'NO',
                message: 'EXPUNGE not implemented'
            });
        }
       
+        // Just return early if not selected
+        if (this.state === 'Authenticated') {
+            return callback(null, {
+                response: 'OK'
+            });
+        }

        // Just unselect if in read only mode
        if (this.selected.readOnly) {
            this.session.selected = this.selected = false;
            this.state = 'Authenticated';
            return callback(null, {
                response: 'OK'
            });
        }
```

titanism:
If nothing selected, then it shouldn't throw an error.  See <https://github.com/nodemailer/wildduck/issues/558> for insight.

### DIFF
--- a/imap-core/lib/commands/close.js
+++ b/imap-core/lib/commands/close.js
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = {
-    state: 'Selected',
+    state: ['Authenticated', 'Selected' ],
 
     handler(command, callback) {
         // Check if EXPUNGE method is set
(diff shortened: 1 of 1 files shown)


andris9:
`CLOSE` is only allowed in Selected state, see [rfc3501#section-6.4](https://datatracker.ietf.org/doc/html/rfc3501#section-6.4)

PR #564, opened 2023-11-26
fix: fixed string conversion to utf8 vs binary (closes #563)

### LINKED ISSUE #563: Bug: Content encoding is incorrect
```eml
Message-ID: <6767f034-bbc1-4086-bced-cd65088eb574@test.net>
Date: Sun, 26 Nov 2023 02:21:11 -0600
MIME-Version: 1.0
User-Agent: Mozilla Thunderbird Beta
From: test <test@test.net>
Subject: test
To: test@test.com
Content-Language: en-US
X-Mozilla-Draft-Info: internal/draft; vcard=0; receipt=0; DSN=0; uuencode=0;
 attachmentreminder=0; deliveryformat=0
X-Identity-Key: id1
Fcc: imap://test%test.net@imap.test.net/Sent Mail
Content-Type: text/html; charset=UTF-8
Content-Transfer-Encoding: 8bit

<!DOCTYPE html>
<html>
  <head>
    <meta http-equiv="content-type" content="text/html; charset=UTF-8">
  </head>
  <body>
    <p>Hey!</p>
    <p>This is a test!</p>
    <p>To show that double line spacing is causing encoding issues with
      WildDuck.  Yo.</p>
    <p><br>
    </p>
    <p>OK<br>
    </p>
  </body>
</html>


```

Rendering incorrectly in Thunderbird:

Screen Shot 2023-11-26 at 2 21 38 AM

When you send the message, the double space is encoded as `ï¿½ `:

Screen Shot 2023-11-26 at 2 28 30 AM

Not sure if this is WildDuck specific, but putting this in here for now until I figure out what is going on.

titanism:
@andris9 this is a core bug that needs merged/fixed

details are in #563 

calling `compiled + "something"` will result in the buffer being converted to utf8 instead of binary, since the default buffer.toString is utf8

### DIFF
--- a/imap-core/lib/imap-composer.js
+++ b/imap-core/lib/imap-composer.js
@@ -60,8 +60,17 @@ class IMAPComposer extends Transform {
             );
         }
 
-        this.push(compiled);
-        this.push('\r\n');
+        // <https://github.com/nodemailer/wildduck/issues/563>
+        // <https://github.com/nodemailer/wildduck/pull/564
+        if (typeof compiled === 'object') {
+          this.push(compiled);
+          this.push('\r\n');
+        } else if (typeof compiled === 'string') {
+          this.push(Buffer.from(compiled + '\r\n', 'binary'));
+        } else {
+          return done(new TypeError('"compiled" was not an object or string'));
+        }
+
         done();
     }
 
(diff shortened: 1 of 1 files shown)


andris9:
If the value is a binary string it gets converted to utf8 with this change which isn’t correct either, there should probably be a conditional check

andris9:
Yes, the easiest check would be to check if the value `typeof` is object, then push, if string, then concat and convert from Binary, and if something else, then maybe return and not push data at all.

andris9 (approved):

PR #568, opened 2023-11-28
fix: FETCH should not compare by length, but by keys

titanism:
This fixes issues such as when message counts in folders were incorrect when messages were moved/appended to them.


andris9:
Could you elaborate a bit here? What exactly was wrong, preferably with a sample IMAP transaction (a list of sample commands and responses, and what is supposed to happen and what actually happens)? FETCH must not return appended messages without prior notification.

PR #588, opened 2023-12-24
Fix openapi.yaml so it passes schema validation

louis-lau:
Openapi spec wasn't passing schema validation, fixed the errors :)

You can check with
```
npx swagger-cli validate docs/api/openapi.yml
```


andris9 (approved):

andris9:
Thanks!

PR #595, opened 2024-01-03
ZMS-117

NickOvt:
Allow to create draft messages with incorrect addresses.
Incorrect addresses are removed, but the message will still be saved, that is if one address is given and it is incorrect the `to` header will be missing, if two addresses are given and one is incorrect the only correct one will be included in the `to` header.

### DIFF
--- a/lib/api/messages.js
+++ b/lib/api/messages.js
@@ -2128,7 +2128,7 @@ module.exports = (db, server, messageHandler, userHandler, storageHandler, setti
             let data = {
                 from: result.value.from || { name: userData.name, address: userData.address },
                 date,
-                to: result.value.to ? result.value.to.filter(toObj => toObj.to !== '') : undefined,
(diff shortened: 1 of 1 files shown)


lib/api/messages.js
```diff
@@ -2121,7 +2128,7 @@ module.exports = (db, server, messageHandler, userHandler, storageHandler, setti
             let data = {
                 from: result.value.from || { name: userData.name, address: userData.address },
                 date,
-                to: result.value.to,
+                to: result.value.to ? result.value.to.filter(toObj => toObj.to !== '') : undefined,
```
andris9, line 2131: `toObj` should have properties `name` and `address`, not `to`, or am I reading this wrong?

andris9 (commented):
See the question about `toObj`

lib/api/messages.js
NickOvt: Yes, good catch, should be address

andris9 (approved):

PR #597, opened 2024-01-04
ZMS-109

NickOvt:
Try to reduce the amount of call to JSON.parse in order to improve the idle CPU performance of wildduck when a lot of connections are open. Also reduce the amount of empty emits by listeners


andris9 (changes requested):

andris9 (approved):

NickOvt:
based on testing (4500 requests each for creating a message (has no payload) and logging out (has payload))
Performance has improved 2x-3x fold in case of no payload that is only .e and no parsing is needed.
Performance is the same or 100ns faster in case when data has the payload.

andris9:
> based on testing (4500 requests each for creating a message (has no payload) and logging out (has payload)) Performance has improved 2x-3x fold in case of no payload that is only .e and no parsing is needed. Performance is the same or 100ns faster in case when data has the payload.

The no-payload version happens way more, as it is triggered any time something happens on the email account, e.g. an email is marked as seen or unseen. So if this use case gets optimized it is a great win 🏆

PR #607, opened 2024-01-22
ZMS-120

NickOvt:
- Added `/health` API endpoint to check the status/health of the WIldduck API

### DIFF
--- a/lib/api/health.js
+++ b/lib/api/health.js
@@ -75,7 +75,8 @@ module.exports = (db, server, loggelf) => {
 
             // 3) test redis PING
             try {
-                await db.redis.ping();
+                // Redis might try to reconnect causing a situation where given ping() command might never return a value, add a fixed timeout
+                await promiseRaceTimeoutWrapper(db.redis.ping(), 10000);
             } catch (err) {
                 loggelf({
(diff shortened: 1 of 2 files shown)


lib/api/health.js
```diff
@@ -0,0 +1,118 @@
...
+                        description: 'Failed',
+                        model: Joi.object({ success: successRes, message: Joi.string().required().description('Error message specifying what went wrong') })
+                    }
+                }
+            }
+        },
+        tools.responseWrapper(async (req, res) => {
+            res.charSet('utf-8');
+
+            const currentTimestamp = Date.now() / 1000;
```
andris9, line 33: This value is basically a float, but you are using it as a string. Maybe round it and then cast it into a string to prevent edge case surprises (eg. if for some value dividing by 1000 does not end up as a fixed number, but something like 12345.99999999...)

```
currentTimestamp = Math.round(Date.now() / 1000).toString();
```

andris9 (changes requested):

NickOvt:
Changed the `db.database.s.client.topology.isConnected()` check to a ping, because even if wildduck was running and then MongoDb got disconnected (manually stopped the process for testing) the `isConnected()` function still returned `true`.

andris9 (approved):

PR #613, opened 2024-02-02
ZMS-119

NickOvt:
- Separate the API docs generation code into a separate public open-source npm package
- Import the package and use it with defined options


NickOvt:
https://github.com/zone-eu/restify-api-generate

https://www.npmjs.com/package/restifyapigenerate

api.js
```diff
@@ -573,14 +575,80 @@ module.exports = done => {
...
-        { path: '/openapi', name: 'openapi-docs-generation' },
-        tools.responseWrapper(async (req, res) => {
-            res.charSet('utf-8');
-
-            const routes = server.router.getRoutes();
-
-            tools.generateAPiDocs(routes);
+    server.pre(
+        restifyApiGenerate(server, {
+            openapiVersion: '3.0.0',
```
andris9, line 580: Could you move this configuration structure into a separate file (json or js) where the config would be loaded from?

package.json
```diff
@@ -95,6 +95,7 @@
         "restify-cors-middleware2": "2.2.1",
         "restify-errors": "8.0.2",
         "restify-logger": "2.0.1",
+        "restifyapigenerate": "^1.1.0",
```
andris9, line 98: Could you remove the caret to keep dependency versions fixed. To upgrade deps, run `npm run update`

andris9 (changes requested):

andris9 (approved):

PR #616, opened 2024-02-06
ZMS-34

NickOvt:
Filter creation is now logged to graylog and authlog.

### DIFF
--- a/lib/api/filters.js
+++ b/lib/api/filters.js
@@ -901,7 +901,10 @@ module.exports = (db, server, userHandler, settingsHandler) => {
                 _user: user,
                 _mailbox: filterData.action.mailbox,
                 _filter_id: filterData._id.toString(),
-                _filter_data: filterData
+                _filter_data: getFilterStrings(filterData),
+                _filter_name: filterData.name,
+                _filter_created: filterData.created,
+                _filter_disabled: filterData.disabled
             });
 
             // Log added filter to authlog as well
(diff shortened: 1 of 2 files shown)


lib/api/filters.js
```diff
@@ -895,6 +895,29 @@ module.exports = (db, server, userHandler, settingsHandler) => {
                 }
             }
 
+            // Log added filter to graylog
+            userHandler.loggelf({
+                short_message: '[FILTERS] Added new filter',
+                _user: user,
+                _mailbox: filterData.action.mailbox,
+                _filter_id: filterData._id.toString(),
+                _filter_data: filterData
```
andris9, line 904: Instead of dumping the entire filter document, it would probably be better to generate a human-readable description with the `getFilterStrings` method https://github.com/nodemailer/wildduck/blob/95f829d16aa24883bd763179581b5288acc51f3d/lib/api/filters.js#L1203

andris9 (changes requested):

lib/api/filters.js
NickOvt: replaced `filterData` with `getFilterStrings(filterData)`. Also added:
```js
_filter_name: filterData.name,
_filter_created: filterData.created,
_filter_disabled: filterData.disabled
 ```
 Because this data is not included in the output of `getFilterStrings()`

andris9 (approved):

PR #639, opened 2024-02-26
ZMS-126

NickOvt:
RestifyApiGenerate now uses the passed Joi package so that version has less impact.
Also __dirname has to be passed in order to maintain correct pathing for the openapidocs.json file


andris9 (changes requested):
Remove the upgrade lock for joi as well:

https://github.com/nodemailer/wildduck/blob/f6f5f5eb65023e8272fbac4d07da1d70070ea5e4/.ncurc.js#L15

andris9 (approved):

PR #655, opened 2024-03-18
ZMS-134

NickOvt:
Downloading an attachment now has an added parameter `sendAsString`, if true, then uses the attachment charset (if available, otherwise utf8) and sends the message as an utf8 string back.
Decoding is piped to the stream

### DIFF
--- a/lib/api/messages.js
+++ b/lib/api/messages.js
@@ -23,6 +23,7 @@ const TaskHandler = require('../task-handler');
 const { prepareSearchFilter, uidRangeStringToQuery } = require('../prepare-search-filter');
 const { getMongoDBQuery /*, getElasticSearchQuery*/ } = require('../search-query');
 //const { getClient } = require('../elasticsearch');
+let iconv = require('iconv-lite');
 
 const BimiHandler = require('../bimi-handler');
 const {
@@ -1505,7 +1506,11 @@ module.exports = (db, server, messageHandler, userHandler, storageHandler, setti
             summary: 'Download Attachment',
             description: 'This method returns attachment file contents in binary form',
             validationObjs: {
-                queryParams: {},
+                queryParams: {
+                    sendAsString: booleanSchema
+                        .default(false)
+                        .description('If true then sends the original attachment back in string format with correct encoding.')
+                },
                 pathParams: {
                     user: userId,
                     mailbox: mailboxId,
@@ -1577,7 +1582,7 @@ module.exports = (db, server, messageHandler, userHandler, storageHandler, setti
                             _id: true,
                             user: true,
                             attachments: true,
-                            'mimeTree.attachmentMap': true
+                            mimeTree: true
                         }
                     }
                 );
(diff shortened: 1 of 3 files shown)


lib/api/messages.js
```diff
@@ -1655,6 +1662,9 @@ module.exports = (db, server, messageHandler, userHandler, storageHandler, setti
             } else if (attachmentData.transferEncoding === 'quoted-printable') {
                 attachmentStream.pipe(new libqp.Decoder()).pipe(res);
             } else {
+                if (!attachmentCharset.toLowerCase().includes('utf') && result.value.sendAsString) {
```
andris9, line 1665: Fails with UTF16. Better check for `/ascii|utf[-_]?8/`

andris9 (changes requested):

lib/api/messages.js
```diff
@@ -1614,6 +1619,8 @@ module.exports = (db, server, messageHandler, userHandler, storageHandler, setti
                 });
             }
 
+            let attachmentCharset = getAttachmentCharset(messageData.mimeTree, attachment)[0];
```
andris9, line 1622: If the response is always an array, then using a destructuring assignment would be more readable

```
const [attachmentCharset] = getAttachmentCharset(messageData.mimeTree, attachment)
```

lib/api/messages.js
```diff
@@ -1655,6 +1662,10 @@ module.exports = (db, server, messageHandler, userHandler, storageHandler, setti
             } else if (attachmentData.transferEncoding === 'quoted-printable') {
                 attachmentStream.pipe(new libqp.Decoder()).pipe(res);
             } else {
+                if (!/ascii|utf[-_]?8/.test(attachmentCharset.toLowerCase()) && result.value.sendAsString) {
```
andris9, line 1665: I you are already using a regex, then it is better to use the case-insensitive flag instead of `toLowerCase()`

```js
!/ascii|utf[-_]?8/i.test(attachmentCharset)
```

lib/api/messages.js
```diff
@@ -3878,3 +3889,25 @@ function parseAddresses(data) {
     walk([].concat(data || []));
     return Array.from(addresses);
 }
+
+function getAttachmentCharset(mimeTree, attachmentId) {
+    if (mimeTree.attachmentId && mimeTree.attachmentId === attachmentId) {
+        // current mimeTree (sub)object has the attachmentId field, and it is the one we search
+        // get the parsedHeader -> content-type -> params -> charset
+
+        return [mimeTree.parsedHeader['content-type'].params.charset || 'UTF-8', true];
```
andris9, line 3898: There is near-zero chance that the `content-type` header is missing, so maybe use optional chaining to prevent exceptions

```js
mimeTree.parsedHeader['content-type']?.params?.charset
```

andris9 (changes requested):

andris9 (approved):

PR #660, opened 2024-03-30
config: have TLS services include tls.toml

msimerson:
1. When I configured `tls.toml`, I expected that to automatically configure TLS for pop3, imap, and lmtp. But it didn't. Populating the cert and key settings within each .toml file did the trick.
2. Based on my readings of the docs for `wild-config`, since there's a `default.toml` file with a `[tls]` section that  includes `tls.toml`, then that should automatically configure the POP3, IMAP, and LMTP daemons TLS settings. Right?
3. If that's how it's supposed to work, any ideas why it wouldn't?
4. I notice that the mongo database has a certs collection. If I POST my certificates there, does that mean I can skip fiddling around with TLS files on the filesystem?
5. I'm thinking the [Setting up TLS](https://github.com/zone-eu/zone-mta/wiki/Setting-up-TLS-or--STARTTLS) wiki page might need some love, and the answers to my questions here would be fine material to include.
6. Why not directly include `tls.toml` in the service configs? It's explicit, and almost self documenting.


msimerson:
From the wildduck TLS [API docs](http://docs.wildduck.email/api/#operation/updateTLSCertificate), "acme, boolean, If true then private key and certificate are managed automatically by ACME"

I have acme.sh set up external (of the container) and it'll be publishing updates of the TLS certificates into the container. That process will be entirely automated, so I'm pretty sure the answer here is `false`? 

Probably even more so false since the ACME servers have no means to reach port 80 within the container and validate a renewal. 

---

I have this little bit of fun added to my acme.sh deploy scripts:

```sh
    # Note: Wildduck only supports RSA certificates (2024-03)
    /usr/local/bin/jo -p -- \
        -s servername=${_cdomain} \
        -b acme=false \
        -s privateKey=@$_ckey \
        -s cert=@$_cfullchain \
        | /usr/local/bin/curl -X POST http://wildduck:8080/certs --json @-
```

I still haven't figured out if/when those certificates will get used.

msimerson:
Thanks for the reply @andris9 .

> Configuring tls.toml should_ configure TLS for IMAP, POP3, and the API. This is the fallback if service-specific TLS settings are not configured. I just tested it with the built-in installer, and it worked—only tls.toml was configured, but the TLS section in the IMAP server config was unset and the cert was correctly selected.

It works for me too. I tried that previously, but I must have had something else going on.

> The TLS certificates from Mongo...are used for SNI only.

Very good, that makes perfect sense. If the TLS request comes in with SNI and a matching certificate is found in MongoDB, then use that certificate. Otherwise, the default TLS certificate specified in the config files is used.

PR #664, opened 2024-04-02
ZMS-142

### LINKED ISSUE #661: Faulty dnsTxt returned on POST /dkim
When making a POST request to the /dkim endpoint, I'm receiving "undefined" in the public key section of the dnsTxt in the response:
  
```
"dnsTxt": {
    "name": "selector._domainkey.domain.com",
    "value": "v=DKIM1;t=s;p=undefined"
}
```

NickOvt:
Fix #661 
Fixed issue that no key der format is created in case no privateKey is given and a pair is generated automatically

### DIFF
--- a/test/api/dkim-test.js
+++ b/test/api/dkim-test.js
@@ -30,6 +30,9 @@ describe('API DKIM', function () {
         expect(response.body.success).to.be.true;
         expect(/^[0-9a-f]{24}$/.test(response.body.id)).to.be.true;
         dkim = response.body.id;
+        expect(response.body.dnsTxt.value).to.not.be.undefined;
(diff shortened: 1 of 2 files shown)


test/api/dkim-test.js
```diff
@@ -16,6 +16,22 @@ describe('API DKIM', function () {
 
     this.timeout(10000); // eslint-disable-line no-invalid-this
 
+    it('should POST /dkim expect success / key empty', async () => {
```
andris9, line 19: This test would have passed even without the changes. Maybe check that `dnsTxt.value` is set and it does not contain any undefined values?

andris9 (changes requested):

andris9 (approved):

PR #672, opened 2024-04-12
ZMS-112

NickOvt:
If a message has the html part then use it for intro instead of text/plain.

### DIFF
--- a/lib/message-handler.js
+++ b/lib/message-handler.js
@@ -318,29 +318,7 @@ class MessageHandler {
                                 ? messageData.text
                                 : messageData.text.substr(0, consts.MAX_PLAINTEXT_CONTENT);
 
-                        messageData.intro = messageData.text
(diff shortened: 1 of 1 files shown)


lib/message-handler.js
```diff
@@ -360,6 +361,30 @@ class MessageHandler {
                                 return html;
                             })
                             .filter(html => html);
+
+                        messageData.intro = htmlToText(messageData.html.join(''))
+                            // assume we get the intro text from first 2 kB
```
andris9, line 366: You are calling this set of regexes, etc, from two different places (both from the plaintext and HTML section), so it might make sense to create a function to convert text input into an intro string instead of defining the same regexes twice.

andris9 (changes requested):

andris9 (approved):

PR #679, opened 2024-04-29
ZMS-147

NickOvt:
`req_end` for `PUT users/:user/autoreply` will not be included in graylog if value is `false`.

### DIFF
--- a/api.js
+++ b/api.js
@@ -101,9 +101,11 @@ const serverOptions = {
             };
 
             Object.keys(req.params || {}).forEach(key => {
-                let value = typeof req.params[key] === 'string' ? req.params[key] : util.inspect(req.params[key], false, 3).toString().trim();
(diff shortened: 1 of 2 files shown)


api.js
```diff
@@ -101,16 +101,24 @@ const serverOptions = {
             };
 
             Object.keys(req.params || {}).forEach(key => {
-                let value = typeof req.params[key] === 'string' ? req.params[key] : util.inspect(req.params[key], false, 3).toString().trim();
+                // cast value to string, as we want to have value present whenever there is some actual value (whatever the data type)
```
andris9, line 104: I think it should be the reverse. If the value is `false`, `null` or empty string, then this key should be ignored and there is no need to log it. This way we can avoid key specific exceptions.

andris9: Ignore this comment, did not realize this PR was already closed. See the same comment in active PR

PR #688, opened 2024-05-09
ZMS-156

NickOvt:
Excluded public and acme endpoints from api docs generation


api.js
```diff
@@ -213,7 +213,7 @@ server.use(
 
 // public files
 server.get(
-    { name: 'public_get', path: '/public/*' },
+    { name: 'public_get', path: '/public/*', exclude: true },
```
andris9, line 216: Use a more descriptive key name than `exclude`. Currently it is hard to understand what is it for.

andris9 (changes requested):

api.js
NickOvt: Renamed to excludeRoute

andris9 (approved):

PR #699, opened 2024-06-27
fix: fixed edge cases when `session` could be `null`

titanism:
If the session is set to `null` and an `err` occurs, for example here:

<https://github.com/titanism/wildduck/blob/c9188b3766b547b091d140a33308b5c3ec3aa1d4/lib/pop3/connection.js#L1010>

Then an uncaught exception will be thrown.

### DIFF
--- a/lib/pop3/connection.js
+++ b/lib/pop3/connection.js
@@ -1,3 +1,4 @@
+
 'use strict';
 
 const crypto = require('crypto');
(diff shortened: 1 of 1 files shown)


andris9 (approved):

lib/pop3/connection.js
```diff
@@ -519,8 +520,8 @@ class POP3Connection extends EventEmitter {
                         _message_id: message.id,
                         _message_uid: message.uid,
                         _mailbox: message.mailbox,
-                        _username: this.session.user && this.session.user.username,
-                        _user: this.session.user && this.session.user.id,
+                        _username: this?.session?.user?.username
```
shaunwarman, line 523:
```suggestion
                        _username: this?.session?.user?.username,
```

Issue #711, opened 2024-07-25
Feat: Add support for `XAPPLEPUSHSERVICE`


titanism:
## Hi folks 👋   We're the team that made **[Forward Email](https://forwardemail.net)**, the only 100% open-source email service out there.

👉 Check it out at https://forwardemail.net!!! 👈 

Keep reading below for insight into how we implemented iOS Push support! ⏬ 

> P.S. Try using coupon code `GITHUB` for a discount 😉 

Ref: <https://docker-mailserver.github.io/docker-mailserver/latest/examples/use-cases/ios-mail-push-support/>

louis-lau:
I agree this would be very nice to have! May not be that easy, as there's no publically available official spec.

louis-lau:
Nice! The fact that people have reverse engineered this before should make things easier. Looking into this was on my list somewhere, but there's about 999 things above it haha. Good luck!

JDENredden:
I can’t contribute anything technically, but I would love to see this implemented. Major quality of life win for iOS users.

JDENredden:
Mailbox.org is another one that has managed to make push email work on iOS. They’re another one to reach out to. Zoho Mail rely on ActiveSync to do push on iOS. On 1 Aug 2024, at 8:39 am, titanism ***@***.***> wrote:﻿
Okay so despite it working and the APN being successfully delivered – it doesn't appear that APPL is accepting it for delivery to the device/account pair.  I've reached out to folks at APPL, Linux (maintainers of that Dovecot project), and the team at Fastmail.  Hoping we can figure out if this is possible still or not, and if not, what are the alternatives.  It seems that Yahoo and Fastmail can only do this approach because of a custom licensing that permits them for com.apple.mobilemail push notifications.

gctwnl:
I've just found this thread (as I ran into this trying my yearly renewal on OS X Server.

I have an OS X Server license (a running Mac mini server with High Sierra and Server.app) but Server.app is no longer capable of renewing the cert. I already have (via MacPorts on macOS) a Dovecot running that does iOS Push. The only thing I need is an updated cert. Is there a script somewhere that can help me with that?

louis-lau:
@gctwnl https://github.com/freswa/dovecot-xaps-daemon/issues/43

gctwnl:
> [@gctwnl](https://github.com/gctwnl) [freswa/dovecot-xaps-daemon#43](https://github.com/freswa/dovecot-xaps-daemon/issues/43)

[...]

PR #712, opened 2024-07-26
feat: added XAPPLEPUSHSERVICE support (per #711)

titanism:
This adds command support for `XAPPLEPUSHSERVICE` with a commented out IMAP command handler showing an example implementation.

### DIFF
--- a/imap-core/lib/commands/xapplepushservice.js
+++ b/imap-core/lib/commands/xapplepushservice.js
@@ -1,0 +1,96 @@
+'use strict';
+
+//
+// Thanks to Forward Email
+// <https://forwardemail.net>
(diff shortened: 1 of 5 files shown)


andris9 (approved):

andris9:
Added the following changes (https://github.com/nodemailer/wildduck/commit/4ef3eaabcad9fcacac5889cf8ad49d4e27233a60):
* fixed CAPABILITY listing for XAPPLEPUSHSERVICE (only visible for authenticated sessions)
* made XAPPLEPUSHSERVICE configurable. Requires `aps.enabled` option to be true, otherwise will reject the command
* updated schema handling (previously invalid input like `tag XAPPLEPUSHSERVICE` with no arguments caused an uncaught exception)

PR #719, opened 2024-08-01
fix: fixed XAPPLEPUSHSERVICE command (needs to return topic)

titanism:
@andris9 this is a core fix for XAPPLEPUSHSERVICE


andris9 (approved):

andris9:
Thanks!

PR #732, opened 2024-09-12
ZMS-169

NickOvt:
1. Mailboxes.js added Joi schema to validate subpath length and count. Values bigger than specified max do not pass validation
2. Added new mailboxes tests accordingly
3. Added same functionality straight in mailboxes-handler.js as well
4. Added IMAP tests accordingly
5. Fix IMAP `RENAME` command small bug, where the specified error code would always be `TEMPFAIL` now it is `err.code || TEMPFAIL`


lib/api/mailboxes.js
```diff
@@ -539,6 +558,24 @@ module.exports = (db, server, mailboxHandler) => {
                 requestBody: {
                     path: Joi.string()
                         .regex(/\/{2,}|\/$/, { invert: true })
+                        .max(MAX_MAILBOX_NAME_LENGTH * MAX_SUB_MAILBOXES + 127)
+                        .custom((value, helpers) => {
+                            const splittedPathArr = value.split('/');
```
andris9, line 563: The same validation function is used in multiple places. Instead, define it as a single validation function in https://github.com/nodemailer/wildduck/blob/master/lib/schemas.js and import it.

andris9 (changes requested):

andris9 (approved):

PR #733, opened 2024-09-13
ZMS-172

NickOvt:
When an attachment gets uploaded calculate sha256 hash of it's content and store in the file metadata in the DB.


lib/attachments/gridstore-storage.js
```diff
@@ -129,6 +130,35 @@ class GridstoreStorage {
...
+                        $set: {
+                            'metadata.fileContentHash': calculatedFileContentHash
+                        }
+                    },
+                    {
+                        returnDocument: 'after'
+                    },
+                    (err, res) => {
+                        if (err) {
+                            return attachmentCallback(err);
```
andris9, line 151: This will invoke `attachmentCallback()` and thus `callback()` twice. Make sure that `callback` is not run until the hash has been updated.

andris9 (changes requested):

lib/attachments/gridstore-storage.js
```diff
@@ -137,6 +139,25 @@ class GridstoreStorage {
...
+                                            'metadata.fileContentHash': calculatedFileContentHash
+                                        }
+                                    },
+                                    {
+                                        returnDocument: 'after'
+                                    }
+                                );
+                            }
+                        }
                         callback(...args);
```
andris9, line 161: There's a potential race condition if we want to use the sha256 hash immediately after the attachment has been uploaded - there's a chance that the `findOneAndUpdate` call is not finished yet, and thus, the value is not available.

andris9 (changes requested):

lib/attachments/gridstore-storage.js
```diff
@@ -33,6 +33,23 @@ class GridstoreStorage {
         });
     }
 
+    updateFileWithContentHashMetadata(args, callback, hash, calculatedFileContentHash) {
```
andris9, line 36: The convention for callbacks is to always use `callback` as the last function argument.

```
updateFileWithContentHashMetadata(args, hash, calculatedFileContentHash, callback)
```

andris9 (changes requested):

PR #742, opened 2024-10-15
ZMS-182

NickOvt:
quering for range can be done as a single document, otherwise  shows error too

### DIFF
--- a/lib/handlers/on-search.js
+++ b/lib/handlers/on-search.js
@@ -220,14 +220,10 @@ module.exports = server => (mailbox, options, session, callback) => {
                                         break;
                                 }
                                 let entry = !op
-                                    ? [
-                                          {
-                                              $gte: value
-                                          },
-                                          {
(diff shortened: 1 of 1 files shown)


lib/handlers/on-search.js
```diff
@@ -279,7 +271,7 @@ module.exports = server => (mailbox, options, session, callback) => {
                                     hdate: !ne
                                         ? entry
                                         : {
-                                              $not: entry
+                                              $not: {}
```
andris9, line 274: This does not seem very logical? 🤔

andris9 (changes requested):

lib/handlers/on-search.js
NickOvt: Oh wow. A an accidental typo. I wonder how it even got pushed, I remember fixing it....
Anyway, next commit will fix it

andris9 (approved):

PR #759, opened 2024-11-28
ZMS-188

NickOvt:
add more readable operationId and response types, fix deps


NickOvt:
@louis-lau Updates the openapi operationIds and responses as requested. Please check with a generator if everything's fine. I tested with `https://openapi-generator.tech/` for `bash` and it looks good to me.

andris9 (approved):

PR #772, opened 2025-01-06
ZMS-196

NickOvt:
when allocating IMAP connection, check if socket is still alive

### DIFF
--- a/lib/imap-notifier.js
+++ b/lib/imap-notifier.js
@@ -393,7 +393,7 @@ class ImapNotifier extends EventEmitter {
         }
 
         let rlkey = 'lim:' + data.service;
-        const { closed: socketClosed, connecting, destroyed } = data.session.socket;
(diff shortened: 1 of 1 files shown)


lib/imap-notifier.js
```diff
@@ -393,19 +393,29 @@ class ImapNotifier extends EventEmitter {
         }
 
         let rlkey = 'lim:' + data.service;
-        this.counters.limitedcounter(rlkey, data.user, 1, data.limit || 15, (err, res) => {
-            if (err) {
-                return callback(err);
-            }
+        const { closed: socketClosed, connecting, destroyed } = data.session.socket;
```
andris9, line 396: Just in case `data.session` is no longer available, use something like `data.session?.socket` instead here

andris9 (changes requested):

andris9 (approved):

PR #779, opened 2025-01-30
ZMS-187: Rewrite addMessage, moveMessage to async/await

NickOvt:
1. Rewrite add message. Add `addAsync` rewrite `add` to use async/await.
2. Rewrite move messages. Add `moveAsync` rewrite `move` to use async/await.

Previously they were fully callback based, which made it harder to read and understand. Current changes should have improved the readibility as it made code more linear.
There are still some function-in-function spots, that may be fixed with later commits and more thorough refactoring.

### DIFF
--- a/lib/message-handler.js
+++ b/lib/message-handler.js
@@ -615,7 +615,7 @@ class MessageHandler {
                             idate: messageData.idate,
                             thread: messageData.thread
                         },
-                        async () => {
+                        () => {
                             // added Entries
                             this.notifier.fire(mailboxData.user);
 
@@ -663,12 +663,7 @@ class MessageHandler {
                             });
 
                             // do not use more suitable finally as it is not supported in Node v8
-                            try {
-                                await processAudits();
-                                resolve(cleanupRes);
-                            } catch {
-                                resolve(cleanupRes);
-                            }
+                            processAudits().then(resolve(cleanupRes)).catch(resolve(cleanupRes));
                         }
                     );
                 });
--- a/lib/message-handler.js
+++ b/lib/message-handler.js
@@ -1378,7 +1378,7 @@ class MessageHandler {
                                             })
                                         );
                                     } catch (err) {
-                                        return reject(err); // in case inner promise rejects just in case reject the outer too
+                                        return reject(err); // inner promise rejects -> just in case reject the outer too
                                     }
                                 } else {
                                     return resolve(updateMessage());
--- a/lib/message-handler.js
+++ b/lib/message-handler.js
@@ -272,7 +272,7 @@ class MessageHandler {
                 prepared = newPrepared; // overwrite top-level original prepared
                 options.maildata = this.indexer.getMaildata(newPrepared.mimeTree); // get new maildata of encrypted message
             } else {
-                // not already encrypted and no need to / cannot ecnrypt
+                // not already encrypted and no need to / cannot encrypt
                 const newPrepared = await this.prepareMessageAsync(options);
                 prepared = newPrepared;
             }
--- a/lib/message-handler.js
+++ b/lib/message-handler.js
@@ -915,451 +915,482 @@ class MessageHandler {
     }
 
     move(options, callback) {
-        this.getMailbox(options.source, (err, mailboxData) => {
-            if (err) {
-                return callback(err);
+        this.moveAsync(options)
+            .then(movedMessageRes => callback(null, ...movedMessageRes))
+            .catch(err => callback(err));
+    }
+
+    async moveAsync(options) {
+        // concurrent promises
+        const [mailboxData, targetData] = await Promise.all([this.getMailboxAsync(options.source), this.getMailboxAsync(options.destination)]);
+
+        const item = await this.database.collection('mailboxes').findOneAndUpdate(
+            {
+                _id: mailboxData._id
+            },
+            {
+                $inc: {
+                    // increase the mailbox modification index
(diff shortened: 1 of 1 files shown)


lib/message-handler.js
```diff
@@ -184,532 +190,538 @@ class MessageHandler {
...
+            throw new Error('Message size ' + options.raw.length + ' bytes is too large');
         }
 
-        this.getMailbox(options, (err, mailboxData) => {
-            if (err) {
-                return callback(err);
-            }
+        // get target mailbox data
+        // get target user data
+        let [mailboxData, userData] = await Promise.all([this.getMailboxAsync(options), this.users.collection('users').findOne({ _id: options.user })]);
```
andris9, line 200: Should verify if userData exists after this operation. getMailboxAsync throws if mailbox is missing but findOne does not

lib/message-handler.js
```diff
@@ -184,532 +190,538 @@ class MessageHandler {
...
-                        }
+                newPrepared.id = prepared.id; // retain original
 
-                        this.attachmentStorage.deleteMany(attachmentIds, maildata.magic, () => callback(...args));
-                    };
+                options.prepared = newPrepared; // new prepared in options just in case
+                prepared = newPrepared; // overwrite top-level original prepared
+                options.maildata = this.indexer.getMaildata(newPrepared.mimeTree); // get new maildata of encrypted message
+            } else {
+                // not already encrypted and no need to / cannot ecnrypt
```
andris9, line 271: Typo

lib/message-handler.js
```diff
@@ -184,532 +190,538 @@ class MessageHandler {
...
+
+        let logTime = messageData.meta.time || new Date();
+        if (typeof logTime === 'number') {
+            logTime = new Date(logTime);
+        }
+
+        let uidValidity = mailboxData.uidValidity;
+        let uid = messageData.uid;
+
+        const finishFunc = // finishFunc:
```
andris9, line 599: finishFunc might be called with an error argument but there is no error handling here

andris9 (changes requested):

lib/message-handler.js
NickOvt: It was neither handled previously. UpdateAddressRegister actually never throws, and thus finishFunc will never actually get any error from it. I removed that part with the upcoming commit

lib/message-handler.js
NickOvt: Where? :)

lib/message-handler.js
```diff
@@ -177,539 +183,545 @@ class MessageHandler {
...
+                            command: 'EXISTS',
+                            uid: messageData.uid,
+                            ignore: options.session && options.session.id,
+                            message: messageData._id,
+                            modseq: messageData.modseq,
+                            unseen: messageData.unseen,
+                            idate: messageData.idate,
+                            thread: messageData.thread
+                        },
+                        async () => {
```
andris9, line 618: addEntries callback can not be async, must be regular callback function. If it throws then will not be caught

andris9 (changes requested):

lib/message-handler.js
```diff
@@ -177,539 +183,540 @@ class MessageHandler {
...
-                            addMessage();
-                        });
-                    }
-                } else {
-                    // message already encrypted
-                    this.prepareMessage(options, (err, newPrepared) => {
-                        if (err) {
-                            return callback(err);
+                            // do not use more suitable finally as it is not supported in Node v8
+                            processAudits().then(resolve(cleanupRes)).catch(resolve(cleanupRes));
```
andris9, line 666: You are triggering `resolve()` here before `processAudits` is finished.

Use this snippet to test function ordering:

```js
function main(){
    return new Promise((resolve, reject)=>{

        let someAction = async () =>{
            await new Promise(r=>setTimeout(r, 100))
            console.log('some action');
        };
        someAction().then(resolve(1)).catch(reject(2))
    });
}
main().then(console.log).catch(console.error);
```

This outputs
```
1
some action
```

andris9 (changes requested):

lib/message-handler.js
```diff
@@ -906,452 +915,483 @@ class MessageHandler {
...
+                            }
+                        });
+                    });
+                });
+            }
+
+            return next();
+        };
+
+        let processNext = async () => {
```
andris9, line 1017: Instead of using a recursive async function `processNext`, use a loop here:

```
let messageData;
while ((messageData = await cursor.next())) {
    try {
        // process message
        ...
    } catch(err){
        ...
    }
}
await cursor.close();
```

This should make the code much more readable

lib/message-handler.js
```diff
@@ -906,452 +915,483 @@ class MessageHandler {
...
+            };
+
+            if (targetData.encryptMessages) {
+                // move target mailbox is encrypted
+                const parsedHeader = (message.mimeTree && message.mimeTree.parsedHeader) || {};
+                const parsedContentType = parsedHeader['content-type'];
+
+                if (parsedContentType && parsedContentType.subtype === 'encrypted') {
+                    // message already encrypted, just continue move
+                    return updateMessage();
```
andris9, line 1270: `updateMessage` is an async function so it is hard to say what exactly is returned here and other places (a Promise). Maybe always await it before returning?

andris9 (changes requested):

andris9 (approved):

PR #782, opened 2025-02-07
fix: fixed byte length comparison causing missing body byte (closes #781)

### LINKED ISSUE #781: Bug: `headerPos` set incorrectly due to varying byte length
Per <https://github.com/nodemailer/wildduck/blob/master/lib/message-splitter.js> the `chunk` might look like:

```
Message-ID: <ui0f6sedxw@blanditiis.io>
To: test@foo.com
List-Unsubscribe: foo@foo.com
From: Test <prani@blanditiis.io>
Subject: testing this
Content-Type: text/plain; charset=us-ascii
Content-Transfer-Encoding: 7bit

H
```

But the raw email sent across actually was:

```
Message-ID: <ui0f6sedxw@blanditiis.io>
To: test@foo.com
List-Unsubscribe: foo@foo.com
From: Test <prani@blanditiis.io>
Subject: testing this
Content-Type: text/plain; charset=us-ascii
Content-Transfer-Encoding: 7bit

Hi,

```

Note that the `i,` gets chopped off, and the resulting body and headers joined is:

```
Message-ID: <ui0f6sedxw@blanditiis.io>
To: test@foo.com
List-Unsubscribe: foo@foo.com
From: Test <prani@blanditiis.io>
Subject: testing this
Content-Type: text/plain; charset=us-ascii
Content-Transfer-Encoding: 7bit

i,
```

e.g. the `H` gets chopped off (first letter after header line break)

The trailing line break is the culprit.  Curious if you think we should dummy-proof this @andris9 (e.g. for developers in case they accidentally send email with a trailing line break)?

I think it'd be nice to dummy-proof against this, perhaps using something similar to this code:

```js
// <https://github.com/sindresorhus/strip-final-newline/blob/a1bfe78e3a3de2f73ed3a7600932d7cc952732b4/index.js#L18-26>
const LF = '\n';
const LF_BINARY = LF.codePointAt(0);
const CR = '\r';
const CR_BINARY = CR.codePointAt(0);

const stripFinalNewlineBinary = (input) =>
  input.at(-1) === LF_BINARY
    ? input.subarray(0, input.at(-2) === CR_BINARY ? -2 : -1)
    : input;
```

Then calling `stripFinalNewlineBinary` perhaps or just looking for it if headers were found.  A rewrite might be needed.

titanism:
ref: <https://github.com/nodemailer/wildduck/issues/781>

### DIFF
--- a/lib/message-splitter.js
+++ b/lib/message-splitter.js
@@ -84,7 +84,7 @@ class MessageSplitter extends Transform {
             this.headerChunks = null;
             this.headers = new Headers(this.rawHeaders);
             this.emit('headers', this.headers);
-            if (data.length - 1 > headerPos) {
+            if (data.length - 1 >= headerPos) {
                 let chunk = data.slice(headerPos);
                 this.bodySize += chunk.length;
                 // this would be the first chunk of data sent downstream


titanism:
bump

NickOvt:
Closed due to the PR not fixing the root cause (fixes only some cases)

PR #806, opened 2025-04-14
ZMS-220: in message handler add bulk_batch_size from settings

NickOvt:


lib/message-handler.js
```diff
@@ -1827,6 +1834,8 @@ class MessageHandler {
                     next();
                 };
 
+                let bulk_batch_size = consts.BULK_BATCH_SIZE;
```
andris9, line 1837: Is this correct?

NickOvt: This is in order to save async calls to settingsManager and only initialize bulk_batch_size from settings once. See lines 1900-1908

andris9 (approved):

PR #822, opened 2025-06-05
ZMS-229: Check for correct mongopaging cursor

NickOvt:
If the only paginated field is `_id` then the cursor is of shorter format. But in any other case (`paginatedField` is anything other than `_id`) the decoded cursor is an array of 2 values. Check for length of decoded cursor and that it is an array in case the `paginatedField` is not an `_id`. If it is `_id` then allow shorter cursor


NickOvt:
In case of the other situation where a longer cursor is given to a pagination with only _id as the paginatedField. The response is 0 results and no nextCursor.

PR #824, opened 2025-06-09
ZMS-238 remove callback based deleteMany from attachment storage. 

NickOvt:
In on-copy.js use async version

### DIFF
--- a/lib/handlers/on-copy.js
+++ b/lib/handlers/on-copy.js
@@ -289,15 +289,18 @@ async function copyHandler(server, messageHandler, connection, mailbox, update, 
                                                     key => prepared.mimeTree.attachmentMap[key]
                                                 );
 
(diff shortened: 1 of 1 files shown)


lib/handlers/on-copy.js
```diff
@@ -289,7 +289,7 @@ async function copyHandler(server, messageHandler, connection, mailbox, update,
                                                     key => prepared.mimeTree.attachmentMap[key]
                                                 );
 
-                                                messageHandler.attachmentStorage.deleteMany(attachmentIds, maildata.magic, () => {});
+                                                messageHandler.attachmentStorage.deleteManyAsync(attachmentIds, maildata.magic);
```
andris9, line 292: Add `.catch()` as well, otherwise if the Promise throws, then nothing will catch it.

andris9 (changes requested):

PR #833, opened 2025-06-19
Add plugin API endpoint support for WildDuck

taxilian:
I'm building a MCP (model context protocol) server for wildduck to allow using an LLM to interact with email accounts; to do that I need to be able to register APIs, thus the PR =]

### DIFF
--- a/api.js
+++ b/api.js
@@ -14,17 +14,15 @@ const ImapNotifier = require('./lib/imap-notifier');
 const db = require('./lib/db');
 const certs = require('./lib/certs');
 const tools = require('./lib/tools');
-const consts = require('./lib/consts');
 const plugins = require('./lib/plugins');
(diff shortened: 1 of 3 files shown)


taxilian:
I ended up expanding this a bit, since a plugin may need to handle authentication differently. I wanted to still be able to use token authentication if needed, though. My endpoint needs to put the token in the URL, since there isn't another good way to do it and that's saved privately. I'm open to counter suggestions

andris9:
While I like that you moved authentication logic into a separate file, this still seems like a huge security risk if /plugin/ paths do not require authentication and operate in the root role in the same context as the regular API. I think you should consider doing this in a separate layer, for example, build a proxy server that passes regular API calls to WildDuck API, but for your plugins, does something else.

PR #844, opened 2025-07-25
fix: fixed on store handler to return `modified` in callback

titanism:
this is a core bug fix, please merge @andris9 @NickOvt 

also added some tests


NickOvt:
Am I missing something or is this already done here: https://github.com/zone-eu/wildduck/pull/844/files#diff-5f6d226ed7f9244b43134d16587a2851f4161d595df5f963bbec2565fb340a1fR138-R140
?
That is, I do not see a point in explicitly adding the arguments, besides for readibility improvements

titanism:
you're totally right 🤦

PR #845, opened 2025-07-30
ZMS-247: implement domain cache to store all used domains of users

NickOvt:


lib/api/addresses.js
```diff
@@ -532,6 +532,7 @@ module.exports = (db, server, userHandler, settingsHandler) => {
             // insert alias address to email address registry
             try {
                 r = await db.users.collection('addresses').insertOne(addressData);
+                db.database.collection('domaincache').insertOne({ domain: addressData.address.split('@')[1] });
```
andris9, line 535: No `await` and no `catch()` handler for the Promise? This would be an uncaught exception if it fails

andris9 (changes requested):

andris9 (approved):

PR #864, opened 2025-09-08
ZMS-264: Add feature to check password with PwnedPassword API on User Login

NickOvt:


lib/user-handler.js
```diff
@@ -432,6 +432,59 @@ class UserHandler {
             return [false, false];
         }
 
+        let isPasswordPwned = false;
+
+        if (config.pwned && config.pwned.enabled && config.pwned.type && config.pwned.type !== 'none') {
+            try {
+                isPasswordPwned = this.checkPwnedPasswordForUser(password);
```
andris9, line 439: checkPwnedPasswordForUser ias async, so this always evaluates to true.

lib/user-handler.js
```diff
@@ -1002,13 +1059,19 @@ class UserHandler {
...
                     user: userData._id,
                     username: userData.username,
                     scope: requiredScope,
                     asp: asp._id.toString(),
                     require2fa: false // application scope never requires 2FA
-                });
+                };
+
+                if (isPasswordPwned) {
+                    authResponse.passwordPwned = isPasswordPwned;
```
andris9, line 1071: Should this also be added to the authentication response schema? https://docs.wildduck.email/docs/wildduck-api/authenticate#responses

lib/tools.js
```diff
@@ -597,6 +598,88 @@ function buildCertChain(cert, ca) {
...
+        return shasum.digest('hex');
+    }
+
+    function get(hashedPasswordPrefix) {
+        const opts = {
+            timeout: API_TIMEOUT
+        };
+
+        return new Promise((resolve, reject) => {
+            const req = https
```
andris9, line 620: Why not use the already included pwndpasswords library? https://github.com/zone-eu/wildduck/blob/d8577623b9fc7749d511ad5df6d8ecd4bc3b4e26/lib/api/users.js#L14

andris9 (changes requested):

lib/user-handler.js
```diff
@@ -432,6 +432,59 @@ class UserHandler {
             return [false, false];
         }
 
+        let isPasswordPwned = false;
+
+        if (config.pwned && config.pwned.enabled && config.pwned.type && config.pwned.type !== 'none') {
+            try {
+                isPasswordPwned = await this.checkPwnedPasswordForUser(password);
```
andris9, line 439: If the user has not changed their password, then there is no need to re-check every time they log in. So instead of caching the result in Redis, store the result in the user object in Mongo the first time, and reset the result when account password is changed. IMAP client can create hundreds of logins in a short period of time and there is no need to check the pwned password every time it happens. Also no point in temporary caching, if the password never changes.

andris9 (changes requested):

lib/user-handler.js
louis-lau: New leaks do get added from time to time, though. If you only ever check once after a password changed, it makes no sense at all to check this on login, there would be no point. The only reason to check on login is because a new leak has happened, otherwise you can just check when setting the password initially.

andris9: well, yeah, that's also valid point 🤔 the issue is, that during last 24 hours we had 9 626 141 successful authentications (and this was a regular day), so even with cashing this would be huge number of haveibeenpwnd requests

NickOvt:
@andris9 @louis-lau 
Made some changes to the PR.
Now the API url is configurable so that own service can be used instead of the publci pwnedpasswords api. That is a localhost can be used or anything else using the PwnedPasswordsDownloader.
Removed redis cache as every user will have the password checked once in two weeks. Redis could save some requests to the API by filtering similar passwords, but passwords may be universally distributed as well.

andris9 (approved):

PR #881, opened 2025-10-02
ZMS-275: search-apply allow to delete found messages

NickOvt:


lib/tasks/search-apply.js
```diff
@@ -18,8 +18,11 @@ let run = async (task, data, options) => {
     const user = new ObjectId(data.user);
 
     const action = data.action || {};
+    let updateAction = false;
+
     if (action.moveTo) {
         action.moveTo = new ObjectId(action.moveTo);
+        updateAction = 'move';
```
andris9, line 25: Fix handling when `moveTo` targets the same folder. Currently, if `moveTo` matches the message's current mailbox, the entire action is skipped, discarding any flag updates (`seen`, `flagged`). Instead, treat this as an `update` action and apply the flag changes.

andris9 (changes requested):

PR #935, opened 2025-12-09
ZMSA-47: dkim legacy decipher fixes

NickOvt:
use createDecipheriv instead of deprecated createDecipher, implement openssl evpBytesToKey


andris9 (commented):
The implementation seems correct but would like to see a unit test with a verified legacy encrypted value (generated using an older Node.js version) that is handled the same way using the proposed solution.

NickOvt:
@andris9 added unit tests

PR #946, opened 2025-12-21
fix: ttlcounter is optional for custom implementations (e.g. @forwardemail)

titanism:
This fixes custom implementations on newer versions of WildDuck, since ttlcounter is a new addition.

### DIFF
--- a/lib/limited-fetch.js
+++ b/lib/limited-fetch.js
@@ -28,7 +28,7 @@ class LimitedFetch extends Transform {
         if (typeof this.options.ttlcounter === 'function') {
             this.options.ttlcounter(this.options.key, this.bytes, this.options.maxBytes, false, () => done());
         } else {
-            done();
+            return done();
         }
     }
 }


titanism:
cc @andris9 im upgrading to latest versions of everything and finding some issues like this along the way, sorry for all the issues

titanism:
✅  PR should be ready, I fixed linting and all tests pass @andris9

NickOvt:
Happy New Year! :tada: :santa:

In my opinion such changes should be made in the fork (or clone) and kept in the fork (or clone) rather than added to the main original application. As it makes the application too broad.
What do you think @andris9?

andris9:
Sorry, this response was written by AI, but it looks correct

---

Thanks for the PR @titanism!

After reviewing the changes, I noticed that WildDuck already has a built-in mechanism for this use case: the `skipCounter` option in `LimitedFetch`.

Looking at `lib/limited-fetch.js`:

```javascript
_flush(done) {
    if (!this.options.maxBytes) {
        return done();
    }

    if (this.options.skipCounter) {
        return done();
    }

    this.options.ttlcounter(this.options.key, this.bytes, this.options.maxBytes, false, () => done());
}
```

In your custom implementation, you can simply pass `skipCounter: true` when creating the `LimitedFetch` stream, and it will skip the `ttlcounter` call entirely. This is already being used in `pop3.js:262` for example.

For your Forward Email implementation, instead of omitting `ttlcounter`, you can explicitly set:

```javascript
new LimitedFetch({
    key: 'pdw:' + session.user.id,
    maxBytes: limit,
    skipCounter: true  // explicitly skip rate limiting
});
```

This approach is preferable because:
1. It makes the intent explicit rather than relying on implicit undefined checks
2. It's already supported and tested in the codebase
3. It avoids the risk of silent failures if `ttlcounter` is accidentally omitted when it should be provided

For the POP3 connection handlers, you would need to ensure your custom `onFetchMessage` handler returns a stream with `options.skipCounter: true` set.

Closing this PR in favor of using the existing `skipCounter` mechanism. Let me know if you have questions about implementing it this way!

titanism:
Thanks, we implemented this approach at <https://github.com/forwardemail/forwardemail.net/commit/e8cbb0f3e693110542ae700781f780c74428e543#diff-d3ba32c6e96fc6b182bd8cfeaa2f37375cc1e88375a3857a98a30609133cc85cR14-R83>

PR #954, opened 2026-01-09
ZMS-106: imap-connection notifiations handling remove lock, use queue instead

NickOvt:


imap-core/lib/imap-connection.js
```diff
@@ -656,14 +634,28 @@ class IMAPConnection extends EventEmitter {
...
-                    // append received notifications to the list
-                    conn.selected.notifications = conn.selected.notifications.concat(updates);
+                    if (!conn.selected.notificationQueueKeys) {
+                        conn.selected.notificationQueueKeys = new Set();
+                    }
+
+                    let queued = 0;
+                    for (let update of updates) {
+                        let key = update && update._id ? update._id.toString() : '';
+                        if (!key || (key && conn.selected.notificationQueueKeys.has(key))) {
```
andris9: The additional key check is redundant, should be

```
if (!key || conn.selected.notificationQueueKeys.has(key))
```

andris9 (changes requested):

PR #955, opened 2026-01-13
ZMSA-67: improve UID SEARCH performance by reducing calls to imap-compiler

NickOvt:


andris9:
## PR Review: ZMSA-67 - Improve UID SEARCH Performance

### Verdict: **RECOMMENDED TO MERGE**

This is a straightforward performance optimization with two improvements that maintain RFC 3501 compliance.

---

### Changes Analysis

#### 1. Reduce Object Creation and `walk()` Calls

**Before:** Creates one attribute object per match, each processed by `walk()`:
```javascript
matches.forEach(nr => {
    response.attributes.push({
        type: 'atom',
        value: String(nr)
    });
});
```

**After:** Single attribute with all values joined:
```javascript
response.attributes.push({
    type: 'TEXT',
    value: matches.join(' ')
});
```

**Impact:** For a search returning 10,000 UIDs:
- Before: 10,000 object allocations + 10,000 `walk()` calls
- After: 1 object allocation + 1 `walk()` call

#### 2. O(n) to O(1) Sequence Lookup

**Before:** Linear search for each match:
```javascript
seq = this.selected.uidList.indexOf(nr) + 1;  // O(n) per match
```

**After:** Build Map once, constant-time lookups:
```javascript
let uidIndex = new Map();
for (let i = 0; i < uidList.length; i++) {
    uidIndex.set(uidList[i], i + 1);  // O(n) once
}
// Then O(1) per match:
let seq = uidIndex.get(matches[i]);
```

**Impact:** For mailbox with 50,000 messages and 1,000 matches:
- Before: O(50,000 * 1,000) = O(50M) operations
- After: O(50,000 + 1,000) = O(51K) operations

---

### RFC 3501 Compliance

The SEARCH response format per RFC 3501:
```
search-response = "SEARCH" *(SP nz-number)
```

Both implementations produce identical wire output:
```
* SEARCH 1 2 3 4 5
```

The compiler handles both `type: 'atom'` and `type: 'TEXT'` by appending the value directly. Since the values are numeric (no special characters), the output is byte-identical.

---

### Code Quality

- Clean, readable implementation
- Handles edge cases:
  - Empty `uidList` (defaults to `[]`)
  - Missing sequence numbers (filters with `if (seq)`)
  - Empty result sets (only pushes if `seqList.length`)
- No breaking changes to the public API

---

### Minor Observations

1. The `for` loop style change from `forEach` to indexed `for` is good - avoids closure overhead
2. The Map pre-population is efficient since uidList is already in memory

### Potential Improvement (Optional)

Could add a small early-exit optimization if `matches` is empty, but the current code handles this correctly anyway.

---

### Testing Recommendation

Verify with:
1. Empty search results
2. Large result sets (10K+ UIDs)
3. Non-UID SEARCH (sequence number mode)
4. SEARCH with MODSEQ criteria (ensure MODSEQ still appended correctly)

---

*Review generated with Claude Code*

NickOvt:
@andris9 needs new review - I optimized mongodb query for uid range a bit

PR #963, opened 2026-01-20
ZMSA-70: optimize search query client command parsing

NickOvt:
an UID SEARCH UID command with 700 UIDs given by client and ~250 of which are ranges could take from 18-36 seconds to parse before in WD.
Now it takes no more than 0.2 seconds.


NickOvt:
Before changes: 
image
After changes:
image
The mailbox contains around 400k messages

PR #965, opened 2026-01-21
ZMSA-71: when requesting filters list value also return original query and action field values

NickOvt:
Also when pushing text part to display query then do not remove the 'text' part (was returned as false before)


lib/api/filters.js
```diff
@@ -360,10 +388,38 @@ module.exports = (db, server, userHandler, settingsHandler) => {
                         name: filterData.name,
                         query: descriptions.query,
                         action: descriptions.action,
+                        originalQuery: {},
+                        originalAction: {},
```
andris9: Add to Joi schema as well

PR #992, opened 2026-02-09
ZMS-8: optimize redis fanout strategy for notifications

NickOvt:
This PR implements a more efficient strategy for sending notifications via pub/sub between WD workers. The strategy also works for distributed workers, for example: server1 and server2 each having 8 workers connect to the redis server. The notifications for a selected user will go only to specified workers of specific servers.


NickOvt:
@andris9 no other global notifications are sent via wd_events. The provided strategy is safe. Any other events use BullMQ via a separate channel not related to wd_events or ImapNotifier.

andris9 (approved):

PR #1000, opened 2026-02-17
ZMS-16: implement message deduplication on MX receive

NickOvt:


lib/filter-handler.js
```diff
@@ -709,6 +709,70 @@ class FilterHandler {
             });
         }
 
+        let sourceMessageId = [].concat((prepared.mimeTree && prepared.mimeTree.parsedHeader && prepared.mimeTree.parsedHeader['message-id']) || []).pop();
+        if (sourceMessageId && prepared.msgid && prepared.hdate) {
+            try {
+                let targetMailboxData = await this.messageHandler.getMailboxAsync({
```
andris9: Could we somehow optimize this and not run the mailbox resolver here, can we get this value from cache or somewhere else?

Or maybe run the check here and then add

```
messageOpts.mailbox = targetMailboxData;
```

to prevent `await this.messageHandler.addAsync(messageOpts)` running this check again

andris9 (changes requested):

PR #1004, opened 2026-02-23
fix(docker): Remove --platform=${BUILDPLATFORM} from Dockerfile

martian-str33t:
## Summary

- Removes `--platform=${BUILDPLATFORM}` from the Dockerfile `FROM` instruction, which causes issues when building for amd64 on ARM hosts (e.g. Apple Silicon). Without this flag, Docker correctly builds for the target platform.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=${BUILDPLATFORM} node:lts-alpine
+FROM node:lts-alpine
 ARG TARGETPLATFORM
 ARG TARGETARCH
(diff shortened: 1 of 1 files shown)


NickOvt:
Hello!

What is the exact issue with this?
BUILDPLATFORM was added exactly for so that then building for, for example, amd on an arm system you pull
node arm, which would run the commands faster, but still build for amd at the end.

Cheers

EDIT: I was incorrect, the PR fixes the buildx issue, mergeing. Thank you!

PR #1005, opened 2026-02-23
fix(api): Fix $text search timeout and rename DB_MAX_TIME_MESSAGES_SEARCH

martian-str33t:
## Summary

- Use longer timeout (`DB_MAX_TIME_MESSAGES`, 2min) for `$text` fulltext searches and shorter timeout (`DB_MAX_TIME_FIELD_SEARCH`, 3s) for indexed `search.*` field queries
- Fix retry error variable (`err` → `error`) in the MaxTimeMSExpired catch block
- Fix retry path setting `maxTimeMS` on wrong object (`opts` → `opts.fields`)
- Rename `DB_MAX_TIME_MESSAGES_SEARCH` → `DB_MAX_TIME_FIELD_SEARCH` for clarity

Without this fix, `$text` searches on large mailboxes crash with `MaxTimeMSExpired` after 3 seconds because they were using the short field-search timeout instead of the generous fulltext timeout.

### DIFF
--- a/lib/api/messages.js
+++ b/lib/api/messages.js
@@ -870,6 +870,7 @@ module.exports = (db, server, messageHandler, userHandler, storageHandler, setti
                 query = prepared.query;
             }
 
+            let isTextSearch = filter && filter.$text;
             let total = await getFilteredMessageCount(filter);
             log.verbose('API', 'Searching %s', JSON.stringify(filter));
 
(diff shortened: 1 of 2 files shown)


NickOvt:
https://github.com/zone-eu/wildduck/commit/63811c7787b5e254f0b425d326762d05a20ba935

In this commit I have added the shorter timeout, because there were issues with search on user accounts with lots of mailboxes (over 200), in which case it made sense to not pin the search on a mailbox, if the search took too much time. The logic was as follows: If the search took more time than 3 seconds, then we get the timeout error, and retry search without pinning the mailbox and with a longer timeout (2 mins).
All other changes in this PR are good.
Remove the changes in consts.js and remove this change `maxTimeMS: isTextSearch ? consts.DB_MAX_TIME_MESSAGES : consts.DB_MAX_TIME_FIELD_SEARCH` and I will accept and merge the PR.

Issue #763, opened 2024-12-09
How to Create Label


iaaxindev:
Hi
I've searched all APIs but am unable to create labels. how can i implement the same.

louis-lau:
Labels are currently not something that's implemented in wildduck.

louis-lau:
I've been wanting to implement it as well.

They'd have to be stored as custom IMAP keywords. There would have to be separate endpoints to manage which keywords get exposed as labels in the api. You should be able to filter messages by labels. There's also a need for counters, so we know how many unread emails each label has, this would work similarly to the mailbox unseen counters. 

That's all I've thought out so far. I've been wanting to do this for a long time, but I just don't ever seem to get around to it. I'm not comfortable enough in the wildduck codebase to write a more specific guide than that.

Maybe @andris9 can chime in if everything I said is accurate and aligns with his views of the project, or perhaps he can give more specific pointers than I can.

andris9:
Well, yeah, labels are out of scope. You can use regular message flags as labels and then search all messages that have the specific flag set to emulate labels.

PR #1031, opened 2026-03-22
Support for custom keywords

louis-lau:
This has been a long time coming, finally found the motivation recently :)

This PR adds support for custom keywords (which are just flags). You can use these to represent labels in a webmail (#763), and sync those same labels with for example Thunderbird over IMAP.

Overview:
- Added keywords array to message CRUD
- Added endpoint to compute counters for a keyword, using redis and deltas in the same manner as mailbox counters
- Once that endpoint has been called and the value is in redis, keyword counters will be emitted over sse (`KEYWORD_COUNTERS`) for that keyword. Implemented it this way to avoid computing counters for irrelevant keywords IMAP clients might add
- Keywords also added to sse events related to message creates/updates. `keywords` on `EXISTS`, and `addedKeywords` and `removedKeywords` on `FETCH`
- Added seperate endpoint to compute counters for flagged messages. Given \flagged is a system flag it shouldn't be part of custom keywords, but knowing the total/unread for flagged messages is still very helpful
- Also added SSE for that (`FLAGGED_COUNTER`)
- No changes to IMAP or indexes, this was already fully supported

Some things:
- The current flag update code didn't allow for both deleting and adding flags in the same call, it would for example error if you set/unset both seen and deleted in the same api call. This is a MongoDB limitation, you can't use `$addToSet` and `$pull` in the same query. I switched it over to a simple aggregation pipeline that does allow for this.
- Setting draft set `flagged` in the db for some reason? Assuming that's wrong and fixed
- There were no tests for filter actions, so while I was there I added tests for all existing actions as well

I normally only work on modern TS codebases, so JS plus the old callback style isn't too familiar for me. Let me know if you see any obvious mistakes :)

### DIFF
--- a/docs/api/openapidocs.json
+++ b/docs/api/openapidocs.json
@@ -6,7 +6,7 @@
         "contact": {
             "url": "https://github.com/zone-eu/wildduck"
         },
-        "version": "1.47.1"
+        "version": "1.48.1"
     },
     "servers": [
         {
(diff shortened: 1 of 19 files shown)


NickOvt:
Hello Louis!

Thank you for your PR!

I have a quick question, does this PR implement just keywords (IMAP standard) or labels similar
to those used by gmail and other email clients (not standard IMAP)?

Best regards
Nick

louis-lau:
Just keywords, but that's enough to quite easily implement labels in your webmail. I've implemented labels in our webmail using this :)

Since labels aren't an IMAP server feature, I didn't feel like it was in scope for wildduck, but keywords are.

louis-lau:
Ah something I forgot to say is that I of course added a keyword filter to the message search endpoint. So you can actually list messages for a keyword. That together with counters makes it usable as labels within a webmail

louis-lau:
Bumpiebump

NickOvt:
Might merge next week when I finally get to it :D

NickOvt:
Hello @louis-lau. It is showing me that this branch has conflicts and I cannot seem to fix them. Can you please review what is conflicting exactly and fix that?
Thank you!

louis-lau:
Just rebased to fix conflicts. Haven't tested the result though, I'll do so later, or you can if you'd like in the meantime :)

NickOvt:
@louis-lau Hello!
This branch again has conflicts :D. Please fix when you have time and after that I'll merge it

NickOvt (approved):

PR #1065, opened 2026-05-08
ZMS-49-3: Parse negative literal size, fix literal handling, fix 0 byte literal handling

NickOvt:
Also add tests


andris9 (approved):
LGTM

PR #1093, opened 2026-06-04
ZMS-52: Add prometheus support

NickOvt:


andris9 (changes requested):
# Code Review: PR #1093 — ZMS-52: Add prometheus support

## Overview

This PR adds Prometheus metrics support to WildDuck. It introduces a new central module `lib/metrics.js` (968 lines) built on `prom-client@15.1.3`, exposes a public `GET /metrics` endpoint from the API service, and instruments code paths across the codebase:

- **API** (`api.js`, `lib/tools.js`): request counts/durations, error codes, service-up gauge.
- **Protocols**: IMAP (`imap-command.js`, `imap-server.js`), POP3 (`lib/pop3/*`), LMTP (`lmtp.js`) — connection counts, command counts/durations, recipient/message outcomes.
- **Auth** (`lib/user-handler.js`): authentication attempt outcomes.
- **Storage** (`lib/message-handler.js`): message operation counts/durations and message-size histograms via prototype monkey-patching.
- **Notifications/events** (`lib/imap-notifier.js`, `lib/events.js`): journal entries, publish modes, listener gauges, queued events.
- **Background jobs** (`tasks.js`, `webhooks.js`, `indexer.js`): task processing, BullMQ queue gauges/worker outcomes, indexer change-stream activity.
- **Cluster aggregation**: master-side IPC aggregation via `prom-client` `AggregatorRegistry`, wired in `server.js`.

It also adds docs (`docs/in-depth/operating-wildduck.md`), a test (`test/api/metrics-test.js`), and the `prom-client` dependency.

The overall design is sound and idiomatic for `prom-client`. However, there are **blocking defects** that must be fixed before merge.

---

## Blocking issues

### 1. Function name mismatch — runtime `TypeError` in the indexer

`indexer.js` calls metric helpers that **do not exist** under those names:

- `metrics.recordIndexerChange(...)` — called in 7 places (e.g. around the live-indexing queue, indexing job success/error/missing paths).
- `metrics.setIndexerLockOwner(isCurrentWorker)` — called in `renewLock()`.

But `lib/metrics.js` exports:

- `recordSearchIndexerChange`
- `setSearchIndexerLockOwner`

Every one of these calls will throw `TypeError: metrics.recordIndexerChange is not a function` at runtime. Because the indexer is feature-gated and **disabled in the test config**, the existing/added tests will not catch this — but it will crash in any deployment with ElasticSearch indexing enabled. Either rename the exports to match the call sites or update the call sites. This is the single most important fix.

### 2. `it.only` in the new test will silently skip all other API tests

`test/api/metrics-test.js` uses `it.only(...)`:

```js
it.only('should GET /metrics expect success without access token', async () => {
```

The Grunt `api` target globs `test/**/*-test.js` and runs them in a **single mocha instance** (via `grunt-mocha-test`). Mocha's `.only` is run-global: when present in any loaded file, mocha restricts the entire run to only `.only`-marked tests. This means the full API suite (users, messages, mailboxes, filters, auth, etc.) would be **silently skipped** in CI, masking regressions across the project. Remove `.only`.

### 3. Malformed `package.json` indentation (breaks lint/prettier)

The diff changes the `"scripts"` key indentation from 4 spaces to 8:

```diff
-    "scripts": {
+        "scripts": {
```

On `master` this is correctly indented at 4 spaces. This is an accidental edit that produces invalid formatting; `prettier`/lint will flag it and the `default` Grunt task runs `eslint` first. Revert this line.

### 4. Debug `console.log` left in the test

```js
console.log(response.text);
```

This dumps the entire metrics exposition to test output. Remove it.

---

## Correctness (non-blocking but worth addressing)

- **`MessageHandler` prototype monkey-patching** (`lib/message-handler.js`): instrumentation reassigns `addAsync`, `moveAsync`, `delAsync`, `put`, `update` on the prototype at module load. I verified all five methods exist and that `addAsync` returns `{ status, data }` with `data.size`, so `result.data.size` is correct. The pattern works but is fragile and non-obvious: any future refactor of these signatures, or a second `require` ordering quirk, can silently break instrumentation. A thin wrapper inside the methods (or an explicit `instrument()` call) would be more maintainable and discoverable than reaching into the prototype from module scope. At minimum, add a comment block explaining why this is done here.
- **`update` source label is hardcoded `'unknown'`** — the `update` operation always records `source='unknown'`, which limits the usefulness of that metric. Consider deriving source from args if feasible, or drop the source label for `update`.
- **`secure` label reflects initial socket only**: IMAP `connection.secure` / POP3 `connection.secured` capture TLS state at connect time. STARTTLS/STLS upgrades are not reflected in `wildduck_connections_total{secure=...}`. Acceptable, but worth a one-line comment so the label isn't misread later.
- **Auth `service` label is partly user-controlled**: in the wrapper on `asyncAuthenticate`, `service = meta.protocol || ...`. For the API login path, `protocol` comes from the request body (Joi `default('API')`). It is passed through `cleanLabel` (lowercased, sanitized, capped at 64 chars), so cardinality is bounded, but a caller can still inject arbitrary-ish low-cardinality values. The `result === 'success' | 'fail' | 'ratelimited' | 'error'` logic and the `RateLimitedError` code check are correct (verified against `rateLimitResponse`). Consider constraining `service` to a known allowlist.
- **Cluster master overhead**: the master process imports `metrics`, which triggers `collectDefaultMetrics()` and registers all gauges with `collect()` callbacks on the master too. `aggregatorRegistry.clusterMetrics()` aggregates *worker* registries, not the master's own, so the master's collection is largely wasted work (and its service-up/connection gauges aren't surfaced). Single-process and worker paths are correct. Not harmful, but the master-side `collectDefaultMetrics()` is unnecessary.
- **`collect()` callbacks hit MongoDB/Redis on every scrape**: the `wildduck_tasks` gauge runs a `tasks` aggregation and `wildduck_bullmq_jobs` calls `getJobCounts` per registered queue on every `/metrics` request. Both are wrapped in try/catch (good), but an unauthenticated scraper can drive these queries. See Security below.

---

## Code quality & style

- The module is well-documented with JSDoc on every helper, uses single quotes, 4-space indentation, `arrowParens: 'avoid'`, and no trailing commas — consistent with the project's prettier config.
- The `'use strict'` header, `getMetric()` get-or-create guard against duplicate registration, and `cleanLabel`/`cleanRoute`/`cleanStatus` sanitizers are all good practice for keeping label cardinality bounded.
- `ecmaVersion` is 2020; the optional catch binding (`catch {}`) used in the `collect()` callbacks is ES2019 and fine.
- **Naming inconsistency**: the exports use `searchIndexer*` naming while the consuming code expects `indexer*` (root cause of blocking issue #1). Pick one vocabulary and apply it consistently.
- The conventional-commit title `ZMS-52: Add prometheus support` is missing a type prefix; the project convention is `feat(...): ZMS-52: ...` (e.g. `feat(metrics): ZMS-52: add Prometheus support`).
- **PR body is empty.** Given the breadth (24 files, +1400/-53), a description of the metric surface, the `/metrics` exposure model, and cluster-aggregation behavior would help reviewers and operators.

---

## Scope creep / noisy diff

- `docs/api/openapidocs.json` changes (the `/authenticated` endpoint, `GetIsUserAuthenticatedResponse`, and the `strict2fa` token description) are **unrelated to Prometheus** — that route and the `strict2fa` logic already exist in `lib/api/auth.js` on master. These deltas were almost certainly produced by running `npm run apidoc` and regenerating pre-existing changes. Consider reverting them to keep the PR focused, or land them separately.
- `package-lock.json` contains unrelated churn (`@msgpackr-extract` 3.0.3→3.0.4, `bullmq` sub-deps, `semver`/`uuid`, several `optional: true` flips) beyond the `prom-client` addition. Ideally the lockfile change for this PR should add only `prom-client` and its tree. Worth confirming this churn is intended.

---

## Performance

- Per-request/command overhead is a `process.hrtime()` pair plus a counter `inc` and histogram `observe` — negligible.
- `prom-client` histograms with the chosen bucket sets are reasonable. Label cardinality is the main risk; it is mitigated by `cleanLabel` (64-char cap, sanitization) and `cleanRoute`, but note that `route` uses `req.route.path` (the route *pattern*, not the concrete URL), which is the correct low-cardinality choice — good.
- The `collect()` callbacks doing DB/Redis work on each scrape are the only non-trivial cost; scrape frequency should be controlled by the operator.

---

## Security

- **`/metrics` is intentionally public** (no access-token check; route name allowlisted in the auth middleware). The docs acknowledge this and recommend firewall/reverse-proxy protection, mirroring ZoneMTA. This is a deliberate design choice but should be called out prominently — metrics can leak operational signal (traffic volumes, error rates, queue depths, version), and the on-scrape `collect()` callbacks let an unauthenticated client trigger a MongoDB aggregation over the `tasks` collection and BullMQ `getJobCounts` on every request. Consider: (a) making the endpoint opt-in via config, and/or (b) supporting a bind-address/allowlist or bearer token for `/metrics`.
- Good: metric labels deliberately exclude user IDs, mailbox IDs, message IDs, addresses, and IPs (documented and enforced via the sanitizers) — this avoids leaking PII through metrics.
- The cluster IPC handler validates `message.type`, `worker.isConnected()`, and pending-request IDs, and uses an unref'd 6.5s timeout — reasonable hardening against hung/rogue messages.

---

## Test coverage

- Only one test is added, and it is currently broken (`it.only` + `console.log`). It also only asserts the happy path for `/metrics` content.
- There is **no test for cluster aggregation**, the `collect()` gauges, the auth-attempt wrapper, the message-operation instrumentation, or the indexer instrumentation (which is where the name-mismatch bug hides). Because indexing/clustering are disabled in the test config, the most error-prone code is entirely untested. Consider at least a unit test for `lib/metrics.js` helpers (`cleanLabel`, `cleanRoute`, `normalizeSource`, `cleanStatusClass`) and a smoke test that `require('./lib/metrics')` and calling each exported function with representative args does not throw — that alone would have caught the indexer name mismatch.

---

## Summary of required changes

1. Fix the `recordIndexerChange`/`setIndexerLockOwner` vs `recordSearchIndexerChange`/`setSearchIndexerLockOwner` name mismatch (runtime crash). **(blocking)**
2. Remove `it.only` from `test/api/metrics-test.js` (would skip the whole API suite). **(blocking)**
3. Revert the `package.json` `"scripts"` indentation change. **(blocking)**
4. Remove the `console.log` in the test. **(blocking)**
5. Recommended: add a no-throw smoke/unit test over `lib/metrics.js` exports; restrict/gate the `/metrics` endpoint or document the bind-address requirement more forcefully; revert unrelated `openapidocs.json`/lockfile churn; add a type prefix to the commit title and a PR description; reconsider the prototype monkey-patching in `message-handler.js`.

Key files: `lib/metrics.js`, `indexer.js`, `test/api/metrics-test.js`, `package.json`, `lib/message-handler.js`, `api.js`.

andris9 (commented):
## Code review — Prometheus support

Nicely scoped feature overall: label cardinality is deliberately bounded (route patterns not URLs, no user/message/IP labels), values are sanitized via `cleanLabel`, gauges use `collect()` correctly, and all instrumentation is non-throwing. One correctness bug plus a few design/scope items below.

### 🔴 IMAP active-connection gauge leaks (never decrements)

`imap-core/lib/imap-server.js`:
```js
metrics.connectionStarted('imap', connection.secure);
connection.once('close', () => metrics.connectionClosed('imap'));
```

`IMAPConnection` never emits a `close` event — its only `emit()` is `emit('error')` (imap-connection.js:505), and teardown in `_onClose()` (imap-connection.js:358) deletes from `server.connections` directly rather than emitting. The `this.emit('close')` at imap-server.js:223 is on the **server**, not the connection.

So `connection.once('close', …)` never fires: `wildduck_connections{service="imap"}` climbs on every accept and never comes back down. (`wildduck_connections_total` is fine; only the gauge is broken.) POP3/LMTP are correct because `POP3Connection` genuinely emits `close` and LMTP uses the smtp-server `onClose`.

**Fix:** emit `close` at the end of `IMAPConnection._onClose()`, or call `metrics.connectionClosed('imap')` from within `_onClose()`. A lifecycle test (accept → close → assert gauge back to baseline) would catch this.

### Design / security

- **`/metrics` is unauthenticated with no opt-out.** It's added to the token-skip list, bypassing `config.api.accessToken`, and there's no `config.api.metrics.enabled` flag — always exposed when the API is enabled. The docs warn about it (standard Prometheus posture), but an operator who wants it off has no switch. Suggest a config toggle.
- **Per-scrape DB/Redis load × workers.** `wildduck_tasks` runs a Mongo `$group` over `tasks` and `wildduck_bullmq_jobs` calls `getJobCounts()` on every registered queue, on every scrape. Under cluster aggregation `AggregatorRegistry` triggers `collect()` in *every* worker even though `aggregator: 'first'` keeps only one — so N−1 of those Mongo/Redis calls are thrown away each scrape. Consider a short-TTL cache in `collect()` or gating these to one process.

### Minor

- **Unrelated `docs/api/openapidocs.json` changes** — adds a `/authenticated` endpoint, `GetIsUserAuthenticatedResponse`, and a `strict2fa` wording tweak with no matching code in this PR. Looks like an incidental `apidoc` regeneration; better split out or confirmed intended.
- **`package-lock.json` churn** beyond `prom-client` (msgpackr-extract 3.0.3→3.0.4, bullmq semver/uuid, `optional: true` flips) reads as `npm install` noise rather than part of this change.
- **`indexer.js`** registers `backlog_indexing` via a throwaway `new Queue(...)` that's never closed — inconsistent with `live_indexing`, which reuses the real instance.
- **IMAP command coverage gaps:** the `_nextHandler` continuation path and the "too many bad commands → disconnect" path return without calling `recordMetric`.

### Tests

The one smoke test is good but exercises only the endpoint in single-process mode — nothing covers the connection-lifecycle gauges (where the bug lives), the command/auth recorders, or cluster aggregation.

**Bottom line:** solid feature; please fix the IMAP gauge leak before merge (it publishes a wrong number), decide whether `/metrics` should be config-gated, and clean up the unrelated docs/lockfile churn.

andris9:
Merge conflicts, otherwise LGTM

andris9 (approved):

PR #1110, opened 2026-06-15
ZMS-80: Add collapseThreads options to list messages in a mailbox

NickOvt:


andris9 (changes requested):
# Code Review: PR #1110 — ZMS-80: Add collapseThreads options to list messages in a mailbox

## Overview

This PR adds a `collapseThreads` query option to the mailbox message-listing endpoint (`GET /users/:user/mailboxes/:mailbox/messages`). When enabled, the endpoint returns only a single representative message per thread (the newest for `order=desc`, the oldest for `order=asc`) instead of every message. It also introduces a related `hasDrafts` boolean to the thread-counter output, indicating whether a thread contains at least one draft.

Key changes:

- **`lib/api/messages.js`**
  - Extracts the projection object into a reusable `getMessageListingProjection(metaData, includeHeaders)` helper.
  - Adds `hasDrafts` to `addThreadCountersToMessageList` (via `$max`/`$cond` on `$draft`) and replaces the per-message `Array.find()` lookup with a `Map` (O(n) instead of O(n·m)).
  - Adds `getCollapsedMessageCount(filter)` (distinct-thread count) and `getCollapsedMessagePipeline(filter, sortAscending)` (sort → group-by-thread → `$first` → `$replaceRoot`).
  - Branches the listing logic to use a new `mongopagingAggregateWrapper` when `collapseThreads` is true.
  - Adds `collapseThreads` to the request schema and `hasDrafts` to three response schemas (mailbox listing, search, archived).
- **`lib/mongopaging-find-wrapper.js`**
  - Adds `mongopagingAggregateWrapper` plus helper functions for cursor-based pagination over an aggregation pipeline (`getAggregateCursorData`, `getAggregateCursorString`, `getAggregateCursorMatch`, `getInvalidPagingCursorError`).
- **`test/api/messages-test.js`**
  - One end-to-end test covering collapse with `desc` pagination (next/previous round-trip), `hasDrafts`, `threadMessageCount`, and `asc` ordering.

## Code Quality & Style

- ESLint passes cleanly on all three changed files (`nodemailer` + `prettier` configs). Indentation, single quotes, 160-char width, no trailing commas are all respected.
- Extracting `getMessageListingProjection` is a nice cleanup that removes the inline projection and de-duplicates the header logic.
- Replacing `threadCounts.find(...)` with a `Map` is a genuine, well-targeted performance improvement to the existing thread-counter path.
- The new wrapper functions are readable and self-consistent.

Minor style/consistency notes:

- The PR title and branch (`ZMS-80`) match the conventional-commit ticket style, but the **PR body is empty**. Given the non-trivial pagination logic, a short description of the cursor design and its limitations would help reviewers and future maintainers.
- The new `hasDrafts` description text is copy-pasted across three schema locations. That mirrors the existing pattern for `threadMessageCount`, so it is consistent, but it does propagate the `collapseThreads`-related `hasDrafts` field to the **search** and **archived** endpoints even though those endpoints do **not** support `collapseThreads`. `hasDrafts` is still meaningful there (it is computed by the shared `addThreadCountersToMessageList`), so this is acceptable — just confirm it is intentional and not accidental scope creep.

## Correctness

The core collapse logic is sound:

- The inner `$sort` in `getCollapsedMessagePipeline` (idate direction = order direction) plus `$first` deterministically selects the newest (desc) or oldest (asc) message per thread. Because the representative selection is deterministic across requests, pagination stays stable page-to-page.
- The aggregate wrapper appends its cursor `$match` **after** the collapse pipeline, so the cursor compares against the representative documents' `idate`/`_id` — which is correct, since those fields belong to the representative message.
- The `total` (`getCollapsedMessageCount`, distinct thread count) is consistent with the collapsed listing (one row per thread). The test asserts `total === 3` for 3 threads, which validates this.
- The next/previous/back-to-page-1 round trip is covered by the test and the tie-break on `_id` in both the sort and the cursor `$match` is correct.

Issues / risks worth flagging:

1. **Cursor encoding inconsistency (works today, fragile).** `getAggregateCursorString` builds the inner cursor with `JSON.stringify([isoDate, idString])`, but it is later unwrapped/parsed via the EJSON-based `getCursorFromCursorWrapper` + `getCursorDataFromCursor` (`EJSON.parse`). This only works because both values are serialized to plain strings (ISO date string + hex id string), so `JSON.parse` and `EJSON.parse` produce identical output. The existing `mongopagingFindWrapper` uses EJSON end-to-end to preserve BSON types. The new path silently relies on the paginated field always being a stringifiable `Date`. If `paginatedField` is ever changed to a non-date/non-string value, this will break subtly. Recommend either using EJSON consistently for symmetry, or adding a comment documenting that the inner cursor is intentionally plain-JSON because both values are stringified.

2. **Cross-mode cursor mixing fails asymmetrically.** Collapse cursors are arrays `[date, id]`; non-collapse (find-wrapper) cursors decode to an object `{idate, _id}`.
   - Passing a *non-collapse* cursor into a collapse request: `getAggregateCursorData` requires an array and throws `Invalid paging cursor` → returned as a 500. Clean enough.
   - Passing a *collapse* cursor into a *non-collapse* request: the find wrapper only checks `Array.isArray(cursorData)` (the collapse cursor *is* a 2-element array), so it passes validation and is handed to `MongoPaging.find` as `opts.next`, potentially producing **wrong/garbage results silently** rather than erroring. This is an edge case (client must flip `collapseThreads` mid-pagination while reusing a cursor), but it is a silent-failure path. Consider tagging the cursor with its mode, or at least documenting that cursors are not interchangeable across modes.

3. **`getInvalidPagingCursorError` returns `err.code = 'cursorerr'`, but the route catches and reports it as a generic `InternalDatabaseError` with HTTP 500.** A malformed *client-supplied* cursor is a 4xx condition, not a server error. This mirrors the pre-existing behavior of `mongopagingFindWrapper` (which also throws `cursorerr` into the same catch block), so it is not a regression — but it would be a good opportunity to map `cursorerr` to a 400 in both paths.

4. **`page` number on the previous-cursor edge.** `setPageToMongopagingCursor(previousCursor, currentPage - 1 < 0 ? 0 : currentPage - 1)` can set page to `0`. The non-collapse wrapper has the identical line, so behavior is consistent, but worth confirming clients tolerate `page: 0` / `page: 1` transitions the same way they do for the existing endpoint.

## Project Conventions

- Follows the established route/schema patterns (Joi `validationObjs`, `booleanSchema.default(false)`, `tools.responseWrapper`).
- Reuses `consts.DB_MAX_TIME_MESSAGES` for the aggregation `maxTimeMS`, consistent with the rest of the file.
- The new wrapper is exported from `lib/mongopaging-find-wrapper.js` alongside the existing one — appropriate location.
- One deviation from convention: the codebase already depends on `mongo-cursor-pagination`, which ships a `MongoPaging.aggregate` function. The PR reimplements aggregate cursor pagination by hand rather than wrapping the library (as `mongopagingFindWrapper` wraps `MongoPaging.find`). The manual approach is defensible (full control over the cursor/page-number wrapping), but it duplicates non-trivial cursor logic that the library already provides and tests. Worth a sentence in the PR description explaining why the library helper was not used.

## Performance

This is the area that most deserves attention.

- The collapse path runs **two full thread groupings per page request**: one in `getCollapsedMessageCount` (for `total`) and one in `getCollapsedMessagePipeline` (for the listing). Both must scan and group all messages matching the filter, on every page.
- There is **no covering index for the `$group` by `thread` scoped to `mailbox`**. `indexes.yaml` has `by_idate_reverse_uid {mailbox:1, idate:-1, _id:-1}` (serves the `$match` + first `$sort`) and `user_messages_by_thread {user:1, thread:1}` (keyed on `user`, not `mailbox`). The `$group` is therefore a **blocking, in-memory stage over the entire matched set** for each request. For large mailboxes this is materially more expensive than the indexed, cursor-only non-collapse path, and `allowDiskUse: true` confirms results can spill to disk.
- The cursor `$match` is applied *after* the collapse `$group`, so it cannot prune work at the index level — every page re-collapses the whole mailbox before slicing to `limit + 1`. Pagination does not get cheaper as you page deeper.

Recommendations:
- Document the cost in the schema description (the existing text says "depending on the order argument" but not that it is significantly heavier than a normal listing).
- Consider whether a `{mailbox:1, thread:1, idate:-1}` (or similar) index would let MongoDB use `DISTINCT_SCAN`/`$group` optimizations to avoid full in-memory grouping, and whether the `total` count can be skipped or cached for large mailboxes.
- At minimum, confirm `DB_MAX_TIME_MESSAGES` (2 min) is an acceptable ceiling for the worst case on the largest production mailboxes.

## Test Coverage

- The single test is reasonably thorough for the **happy path**: it builds a thread (root + reply) plus two standalone messages, then asserts collapse `total`, the representative IDs per page, `threadMessageCount`, `hasDrafts` (both true and false), `nextCursor`/`previousCursor` presence, the next→previous→page-1 round trip, and `asc` ordering with the oldest representative.

Gaps worth adding:

- **`includeHeaders=true` with collapse** — the projection branch differs and is untested in collapse mode.
- **`unseen` filter with collapse** — the `filterUnseen` branch feeds into `filter` and the collapse count/pipeline; untested.
- **Invalid / mismatched cursor** behavior in collapse mode (the `Invalid paging cursor` path), and the cross-mode cursor mixing described above.
- **Empty mailbox / single-thread / `limit` exactly equal to thread count** boundary cases (off-by-one on the `limit + 1` extra-row logic).
- **`metaData=true` with collapse** — verifies the `meta` vs `meta.from` projection branch through the aggregate path.

No tests were added for `mongopagingAggregateWrapper` in isolation; all coverage is via the API. A couple of unit tests around the cursor encode/decode helpers would harden the fragile JSON/EJSON interplay noted above.

## Security

- No new authentication/authorization surface; the collapse path runs under the same `roles.can(...).readOwn/readAny('messages')` checks and the same `mailbox`/`user` scoping as the existing listing. The aggregation `$match` reuses the already-validated `filter` (with `mailbox` as an `ObjectId`), so there is no injection vector from user input.
- Cursor inputs are validated by Joi (`nextPageCursorSchema` / `previousPageCursorSchema`) and then defensively parsed; malformed cursors throw rather than executing arbitrary content. The `new ObjectId(id)` reconstruction is wrapped in try/catch.
- `allowDiskUse: true` on an un-indexed full grouping is a mild **DoS consideration**: a user with a very large mailbox can trigger expensive disk-spilling aggregations on every paginated request. It is bounded by `maxTimeMS`, but combined with the double-grouping (count + listing) it is worth keeping in mind for rate-limited or abusive scenarios.

## Specific Suggestions

1. Make the inner cursor encoding consistent with the find wrapper (use EJSON) or add a comment explaining the deliberate plain-JSON choice and its assumption that both cursor values are stringified.
2. Map `err.code === 'cursorerr'` to an HTTP 400 (`InputValidationError`-style) instead of a 500 in the listing handler — ideally for both wrappers, since this is a pre-existing wart.
3. Guard against cross-mode cursor reuse (tag the cursor with its pagination mode, or document clearly that collapse and non-collapse cursors are not interchangeable).
4. Add an index (e.g. `{mailbox:1, thread:1, idate:-1}`) or evaluate one, and update `indexes.yaml`; document the performance cost in the `collapseThreads` schema description.
5. Consider whether `total` (the second full group) can be made optional or cached, to avoid two full groupings per page.
6. Expand tests: `includeHeaders`, `unseen`, `metaData`, invalid-cursor, and boundary (`limit == thread count`) cases.
7. Fill in the PR description with the design rationale (cursor format, why not `MongoPaging.aggregate`, and the performance trade-off).

## Overall Assessment

The feature is well-structured, the collapse semantics are correct, the cursor pagination round-trips properly, and the incidental `Map` cleanup is a nice win. The main concerns are **performance** (un-indexed double `$group` per page on potentially large mailboxes, with `allowDiskUse`) and a few **robustness/consistency** rough edges in the hand-rolled aggregate cursor (JSON vs EJSON, 500-on-bad-cursor, cross-mode cursor mixing). None of these are blockers, but I would want the index/performance question answered and the schema description updated before merge, plus a couple more tests around `includeHeaders`/`unseen`/invalid cursors. ESLint is clean and conventions are followed.

andris9 (changes requested):
Nice work on the aggregate pagination — cursor round-trips, look-ahead, and the empty/single-thread cases are all handled and tested. One blocking regression before merge, plus a cheap robustness fix.

**Blocking — `includeHeaders=<specific keys>` regression on the mailbox listing.** The refactored `getMessageListingProjection()` uses a strict `includeHeaders === true` check, but the listing parses a comma-separated `includeHeaders` into an **array**, which now falls to the `else` branch and only projects the 7 "required" headers. The old inline code used a truthy check, so an array projected `mimeTree.parsedHeader` (all headers). Result: `GET .../messages?includeHeaders=List-ID,X-Foo` used to return those headers and now silently drops them (`formatMessageListing` looks up keys that were never fetched). This isn't covered — the new test only uses `includeHeaders=true`. Fix: treat a non-empty array like `true` (or project the specific `mimeTree.parsedHeader.<key>` paths), and add a test for the specific-header-list case with `collapseThreads` on and off.

**Should fix — `getCollapsedMessageCount` is missing `allowDiskUse: true`.** The listing pipeline has it but the count pipeline doesn't; its `$group` by thread can hit the 100MB limit on a large mailbox and 500 the whole request (count runs before the listing). One-line parity fix.

Non-blocking follow-ups:
- `hasDrafts` means "thread has a draft" in collapsed/archived responses but "a draft is on this page" in non-collapsed ones, and can flip between pages. It's documented, but consider using the accurate per-thread aggregate value everywhere for consistency.
- Duplicate `_id` keys in the cursor literals when `paginatedField === '_id'` (reachable via search with no `order`) — works via last-key-wins, but please special-case it explicitly.
- Worth a note that collapsed vs non-collapsed cursors aren't interchangeable, and that `collapseThreads` (not just `threadCounters`) carries query overhead.

andris9 (approved):

PR #1111, opened 2026-06-17
ZMS-77: Emit marked.ham webhook when starring or explicitly marking ham a message

NickOvt:


andris9 (changes requested):
# Code Review: PR #1111 — ZMS-77: Emit `marked.ham` webhook when starring or explicitly marking ham a message

## Overview

This PR adds the ability to emit the `marked.ham` webhook event when a message is starred (`flagged=true`) or explicitly marked as ham via a new `markHam` request parameter, **without requiring the message to be moved out of the Junk folder**. Previously, `marked.ham` / `marked.spam` events were only emitted as a side effect of moving messages into/out of the Junk folder (via the `entry.junk` signal on `EXISTS` journal entries).

The change spans three files:

1. **`lib/api/messages.js`** — Adds a `markHam` boolean request-body field (default `false`) to both `PUT /messages` endpoints (path-param and body variants).
2. **`lib/message-handler.js`** — Refactors `MessageHandler.update()` so it can run for a "no flag change but markHam" case, publishes `MARKED_HAM` events directly (batched) for matched messages, and tags move-generated `EXISTS` entries with `entry.markHam` when the move carried `flagged`/`markHam` updates.
3. **`lib/imap-notifier.js`** — Extends `pushToJournal()` so `EXISTS` entries flagged with `markHam` also publish a `marked.ham` event, and disambiguates the SPAM-vs-HAM decision.

It also includes two incidental prettier reformattings (lines collapsed/expanded in the encrypted-move quota update and the attachment-release log call).

## Code Quality and Style

- **Prettier/ESLint conformance:** The incidental reformatting at the encrypted-move `updateQuotaAsync` call and the `log.error('MOVE', ...)` call are legitimate prettier fixups (collapsing to a single line vs. multi-line argument wrapping to respect the 160-char width). These are fine, though strictly speaking they are unrelated to ZMS-77 and add review noise. Consider keeping unrelated formatting churn out of feature PRs.
- New code uses the project conventions (4-space indent, single quotes, no trailing commas, `let`). Nothing jumps out as a style violation.
- The refactor of `update()` is sizable (the method roughly doubles in length). The introduction of named helper closures (`processMailbox`, `publishMarkHamEntries`, `flushUpdates`, `markMessageHam`) is a reasonable way to share logic between the two branches, but the method is now quite dense and carries three interleaved concerns (MODSEQ acquisition, per-message flag updates, and ham publishing). A short block comment explaining the `!update` vs `update` split would help future maintainers.

## Correctness

The core logic appears correct. Key paths I traced:

- **`markHam` computation:** `let markHam = changes && (changes.markHam === true || changes.flagged === true)`. Since the API schema defaults `markHam` to `false`, a plain flag update (e.g. `seen=true`) won't trigger ham publishing — good.
- **`!update && !markHam` guard:** The early `Nothing was changed` return now correctly allows a `markHam`-only request to proceed.
- **`markHam`-only branch:** Uses `findOne` (no MODSEQ increment, no per-message `findOneAndUpdate`), increments `updatedCount`, and batches `markHamEntries`. The `findOne` result is wrapped as `{ value: mailboxData }` so it reuses `processMailbox`'s `item.value` contract — a clean adaptation.
- **`flagged=true` path:** `update` is `true` (flagged sets it) AND `markHam` is `true`, so messages go through the normal `findOneAndUpdate` update path and each updated message is additionally pushed to `markHamEntries` exactly once. No double publish.
- **`imap-notifier.js` disambiguation:** `ev: entry.command === 'EXISTS' && entry.junk > 0 ? MARKED_SPAM : MARKED_HAM`. This preserves the prior semantics: an `EXISTS` entry with `junk > 0` (moved into Junk) → SPAM; `junk < 0` (moved out of Junk) → HAM; and now a `markHam`-tagged entry → HAM. Correct.

**Issues / risks to confirm:**

1. **`entry.user`/`entry.mailbox` presence for `markHam`-only EXISTS entries (potential crash).** In `pushToJournal()`, the publish call dereferences `entry.user.toString()`, `entry.mailbox.toString()`, `entry.message.toString()`. For the existing `junk` path these fields are guaranteed because the entry is an `EXISTS` move entry that flows through the `entries.forEach` block setting `entry.user`/`entry.mailbox` (see `addEntries` ~lines 281–286). The new `entry.markHam` tag is only added in `updateMessage` (the move helper), which flows through the same `addEntries` enrichment, so `user`/`mailbox` should be populated. This is fine for the move path, **but** it is worth an explicit confirmation that no caller can ever enqueue a `markHam` entry that bypasses that enrichment (e.g. an entry that already has `modseq` set, which skips the `findOneAndUpdate` mailbox lookup but still passes through the `forEach`). The `forEach` always runs, so `entry.user` is set regardless — looks safe, but this is the most fragile coupling in the PR and deserves a test.

2. **Two separate publish mechanisms for the same event.** `marked.ham` is now published in two places with slightly different shapes:
   - `message-handler.js` `update()` → `publish(this.redis, { ev: MARKED_HAM, user: mailboxData.user, mailbox: mailboxData._id, message: entry.message, id: entry.uid })`
   - `imap-notifier.js` `pushToJournal()` → `publish(this.redis, { ev: ..., user: entry.user.toString(), mailbox: ..., message: ..., id: entry.uid })`

   In the `update()` path the `user`/`mailbox`/`message` values are raw `ObjectId`s, whereas the notifier path calls `.toString()`. This is harmless because `events.publish()` normalizes any value with a `toHexString()` method to a string — but the inconsistency is easy to trip over later. Consider normalizing one way for readability.

3. **`put()` not covered.** `MessageHandler.put()` (a separate single-message store path, ~line 1402) also creates `junk`-tagged `EXISTS` entries but was not given `markHam` handling. This is likely intentional (`put()` doesn't take `flagged`/`updates`), but if any flow relies on `put()` for moves-with-flag-changes, ham emission would be missed. Worth a one-line confirmation that `put()` is out of scope.

4. **Ordering of journal write vs. event publish unchanged.** In `pushToJournal()`, events are still published *before* the `journal.insertMany`. If the insert later fails, the webhook has already fired. This is pre-existing behavior, not introduced here, but the PR widens the set of events affected by it.

## Performance

- The `markHam`-only branch avoids the per-message `findOneAndUpdate` write entirely (only a cursor scan + batched publishes), which is efficient.
- Batching respects `consts.BULK_BATCH_SIZE` / the configurable `const:max:bulk_batch_size`, consistent with the existing notify-flush cadence. `flushUpdates` flushes when **either** `notifyEntries` or `markHamEntries` reach the batch size, which is correct.
- `publishMarkHamEntries()` uses `Promise.all` to fan out publishes per batch — reasonable. For very large ham operations this enqueues one BullMQ job per message, which matches the existing per-message event model (no regression), but be aware a "mark 100k messages ham" operation will enqueue 100k webhook jobs. That is inherent to the feature, not a defect.

## Project Conventions

- API additions follow the established pattern: Joi `booleanSchema.default(false).description(...)` in `validationObjs.requestBody`, applied identically to both `PUT` variants. Good.
- The schema description (`'... Setting flagged=true also emits marked.ham'`) correctly documents the coupled behavior — helpful and accurate.
- Event constant reuse (`MARKED_HAM` from `lib/events.js`) is consistent with the codebase.
- The new request field is **not** reflected in the response schema or in any `apidoc`/generated docs beyond the inline description — acceptable since it's input-only, but confirm the published API docs regenerate.

## Test Coverage

- **No tests were added.** `grep` across `test/` finds zero references to `markHam`, `MARKED_HAM`, or `marked.ham`, and no webhook assertions exist for this flow. Given this PR introduces a new public API parameter and a new event-emission path (including the fragile `entry.user`/`entry.mailbox` coupling in `pushToJournal`), this is the most significant gap. Recommended additions:
  - `PUT /messages` with `markHam=true` (no other changes) returns success with a non-zero `updated` count.
  - `PUT /messages` with `flagged=true` emits `marked.ham`.
  - `markHam=true` does not error when nothing else changed (previously would have returned `Nothing was changed`).
  - A move-into-Junk still emits `marked.spam` and move-out-of-Junk still emits `marked.ham` (regression guard for the `imap-notifier.js` ternary change).

## Security

- The new `markHam` field is a plain boolean validated by Joi; no injection surface.
- Permission checks are unchanged — the existing `updateOwn('messages')` / `updateAny('messages')` gating in `putMessageHandler` still applies, so `markHam` cannot be triggered without message-update rights. Good.
- No new data is exposed in responses. Event payloads contain only IDs already available to webhook subscribers.

## Specific Suggestions

1. **Add tests** for the four scenarios above — this is the highest-value follow-up.
2. **Unify the publish payload shape** between `update()` and `pushToJournal()` (either both `.toString()` or both raw ObjectIds) to reduce confusion.
3. **Split out the unrelated prettier reformatting** (encrypted-move quota call and attachment-release log) into a separate commit/PR, or at least note it in the PR description, to keep the feature diff focused.
4. **Add a brief comment** in `update()` explaining the `!update`/`markHam`-only branch and why it uses `findOne` instead of the MODSEQ-incrementing `findOneAndUpdate`.
5. **Confirm `put()` scope** — explicitly note whether the lack of `markHam` propagation in `put()` is intentional.
6. **PR hygiene:** the PR body is empty. Add a description linking ZMS-77 and summarizing the new `markHam` parameter and the `flagged⇒marked.ham` coupling so reviewers and changelog tooling have context. The branch/commit message follows the conventional-commit ticket-ref convention.

## Risk Assessment

**Overall: low-to-moderate.** The logic is sound and backward-compatible (existing junk-move SPAM/HAM emission is preserved by the refactored ternary). The main residual risks are (a) the absence of tests around a new event path, and (b) the implicit dependency that every `markHam`-tagged `EXISTS` entry passes through `addEntries`' field-enrichment before `pushToJournal` dereferences `entry.user`/`entry.mailbox`. Both are addressable with targeted tests. I'd recommend adding test coverage before merge.

PR #1122, opened 2026-07-16
ZMS-70: Change auth ratelimit order

NickOvt:
This PR fixes a situation where in case a user changes their master password and then the client bombards the server with the old master password the user will not be completely blocked from logging into their account.
Order of ratelimits:
1. Check ip ratelimit
2. Check asp ratelimit
3. Check account ratelimit (master password)

This ensures that even if master password login is ratelimited it is still possible to login at least with an ASP

### DIFF
--- a/lib/user-handler.js
+++ b/lib/user-handler.js
@@ -635,6 +635,7 @@ class UserHandler {
 
         let appPassword;
         let appPasswordError;
+        let appPasswordScopeMismatch;
         let appPasswordFailure = 'Invalid Auth - Master password did not match';
 
         if (requiredScope !== 'master') {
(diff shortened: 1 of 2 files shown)


andris9:
One blocker: the ASP scope-mismatch path loses its audit metadata.

Previously `meta.asp`, `meta.aname`, `passwordType` and `passwordId` were all set *before* the scope check, so a rejection produced an authlog entry naming the offending ASP and an `[AUTHFAIL]` tagged `_password_type: 'asp'`. In the new code the error is stashed without assigning any of them (`lib/user-handler.js:687-694`), and the deferred handler only sets `meta.source = 'asp'` (`:1113-1119`). Since `logAuthEvent` persists the whole `meta` object, the authlog no longer records *which* ASP was rejected, and the outer catch logs `_password_type: 'master'` / `_password_id: undefined` — mislabeling an ASP failure as a master password failure.

Fix is to set those four when stashing the `InvalidAuthScope` error.

andris9 (approved):

PR #1146, opened 2026-08-21
fix list-unsubscribe header parser for get message endpoints

NickOvt:


andris9 (approved):
Pushed follow-up fixes on top: salvage valid URIs from partially malformed values, any URI scheme per RFC 2369, unparseable values now go to name so address always holds a valid URI, List-ID parsing moved to the same module, and a punycode crash guard in decodeAddresses. Unit and integration suites pass locally.

PR #1150, opened 2026-08-25
fix(api-put-queue): ZMS-87 add endpoint to update message in queue

NickOvt:
Adds `PUT /users/{user}/outbound/{queueId}` for changing the delivery time of an outbound message that is still queued. Delivery times in the past are replaced with the current time, and times past the queue expiry are rejected.

For a message submitted by the user the Date header is kept in sync with the new delivery time, in the queued headers and in every stored representation of the message. Forwards and autoreplies are rescheduled but keep the Date value of the original message, and the response reports which of the two happened.

### Notes

- All remaining queue entries of a message have to be reschedulable. The Date header is shared by every recipient, so rescheduling only the unlocked entries would deliver the locked ones with a Date value from the future. The request fails with `QueueEntryLocked` in that case, reporting how many entries are being delivered.
- Rewriting the Date header changes the size of the stored message, so `size` is recomputed and the difference is applied to the user quota. IMAP serves `RFC822.SIZE` from the stored value but rebuilds the `BODY[]` literal from the mime tree, so these have to agree.
- Ownership of a queue entry is taken from the queue metadata, and falls back to the `outbound` reference on the user's own messages. Not every producer stamps the owner (autoreplies did not, `haraka-plugin-wildduck` still does not for forwards) and entries queued before a producer is fixed never will, but the `outbound` reference is how the user got the queue ID in the first place. Entries owned by another user are rejected either way.
- `sender.maxQueueTime` is new in `config/sender.toml` and must match the `maxQueueTime` value of ZoneMTA. The MTA expires queue entries by insertion time and drops them without a bounce, so a delivery time past that point would silently lose the message.
- The regenerated OpenAPI docs also pick up `includeHasDrafts`, `collapseThreads` and `markHam`, which landed on master without a regeneration.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -52,6 +52,7 @@ module.exports = function (grunt) {
                     'test/filtering-tools-test.js',
                     'test/hibp-tools-test.js',
                     'test/list-headers-test.js',
+                    'test/maildropper-test.js',
                     'test/tools-test.js'
                 ]
             },
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -53,6 +53,8 @@ module.exports = function (grunt) {
                     'test/filtering-tools-test.js',
                     'test/hibp-tools-test.js',
                     'test/list-headers-test.js',
+                    'test/maildropper-test.js',
+                    'test/message-handler-update-test.js',
                     'test/metrics-config-test.js',
                     'test/prometheus-test.js',
                     'test/tools-test.js'
--- a/config/sender.toml
+++ b/config/sender.toml
@@ -13,3 +13,8 @@ collection="zone-queue"
 # Must be shared with haraka-plugin-wildduck
 # If not set then looping is not tracked
 #loopSecret="secret value"
+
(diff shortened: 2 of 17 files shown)


andris9 (approved):
Approved. Fixes for the ownership check, the Date header drift and the response semantics are pushed to the branch (a05d06a). Not merging, leaving that to you.

andris9 (approved):
Reviewed again after 44f2e2a1 and pushed the fixes there.

Main one: rescheduling only the unlocked entries still rewrote the shared Date header, so the recipients that were mid-delivery went out with a Date value from the future. The update now refuses unless every remaining entry could be rescheduled.

Also fixed: the stored `size` was not recomputed when the Date header changed, so `RFC822.SIZE` disagreed with the `BODY[]` literal; the ES index kept the old date and size; a failure in the stored-copy sync answered 200 with a bare `{error}`; and the tightened ownership check made autoreplies and haraka-created forwards impossible to cancel, which now falls back to the `outbound` reference on the user's own messages.

Full suite green locally: 49 API, 137 server-backed, 458 IMAP, 187 unit.

PR #1151, opened 2026-08-25
feat(prometheus): ZMS-93: Add prometheus as separate service so it is separate from general API

NickOvt:


andris9 (approved):
Approved. Pushed 85e416dd on top, which fixes the review findings:

- `[api.metrics]` is still honored as a deprecated fallback, otherwise every deployment that enabled metrics in 1.49.6/1.50.0 would silently stop exposing them on upgrade (their `/etc/wildduck` copy has no `metrics.toml`, so `config.metrics.enabled` is false).
- The HTTPS metrics listener now registers for cert reloads. `lib/certs.js` falls back to `setSecureContext` for node core servers and keeps the options the server was created with, so `[metrics.tls]` hardening survives a renewal.
- Metrics listener starts first, so a later service failing to bind is visible as `wildduck_service_up{service="imap"} 0` instead of a scrape failure. Verified end to end.
- `HEAD` and a trailing slash on `/metrics` are accepted, other methods get 405 + `Allow`.
- Dropped `wildduck_service_up{service="metrics"}` (it could only ever be 1) and the unused `createServer`/`metricSource` export.
- `worker.js` startup is a table now instead of nine nested callbacks.
- Tests: the default-disabled guard is scoped to the top-level keys again, the tautological `config.api.enabled` assertions and their global config mutation are gone, and there is coverage for the legacy config fallback, HEAD, trailing slash, 405, the served certificate and the reload registration.

Not merging, that is yours.

PR #1153, opened 2026-08-28
ZMS-99: when updating message with collapseThreads true also update all messages in thread

NickOvt:


andris9 (approved):
Reviewed and pushed fixes for the issues found (41551b16). Summary of what changed:

- **DELETE no longer aborts mid-thread.** Each message is deleted in its own try/catch and the response carries `deleted`/`errors` counts. Previously one failure part-way through a thread returned a bare 500 after some messages were already permanently gone, with no indication of how many.
- **Drafts pulled in by the thread expansion are archived.** `archive: !flags.includes('\\Draft')` meant a draft reply in the thread was hard-deleted with no restore path (`lib/tasks/restore.js` only reads the `archived` collection). A draft the request points at directly is still not archived, as before.
- **`metaData`, `expires` and `draft` are rejected with `updateThread`.** These are per-message values; fanning them out overwrites data on siblings that were never part of the request.
- **`maxTimeMS` on both expansion queries and the delete cursor**, matching the rest of the module.
- **The expanded uid list goes through `tools.checkRangeQuery`** instead of a hand-rolled ``, so contiguous thread siblings collapse to an index range rather than N point seeks.
- **The mailbox is resolved once** instead of once per deleted message, and an unknown mailbox returns `NoSuchMailbox` like the sibling endpoint. Error responses now carry a `code`.

Tests added for `moveTo` + `updateThread` (previously untested), draft archiving on both paths, and the new validation/not-found responses.

Not changed, worth knowing: the delete loop is now a third copy of the same cursor+`delAsync` walk (`deleteMessagesInMailbox`, `lib/tasks/clear-folder.js`) and could be factored into one helper, and neither this loop nor its siblings batch quota updates and notifications the way `on-expunge` does with `delayNotifications`. Both are pre-existing and belong in their own change.

PR #1158, opened 2026-09-02
fix(proxy): guard the raw socket while parsing the PROXY header

metrotyranno:
## Summary

When the PROXY protocol is enabled (`imap.useProxy`), a client that drops the
connection while the server is still reading the PROXY header crashes the whole
process with an unhandled `ECONNRESET`. The PROXY header is read before
authentication, so any client that can reach a PROXY-fronted listener can
trigger it — an unauthenticated denial of service. Both the IMAP server
(`imap-core`) and the POP3 server are affected; they carry the same
`_handleProxy` implementation.

## Root cause

`_handleProxy` parses the PROXY header directly on the freshly accepted socket,
before the connection object (and its error handler) exists. While waiting for
the header it attaches only a `readable` listener:

```js
socket.on('readable', socketReader);
```

There is no `error` (nor `close`/`end`) handler on the socket during that
window, so a socket error emitted there has no listener and is thrown, taking
the process down.

Two ways to hit it:

1. **Reset while waiting for the header** — the client sends `RST` before/while
   the header arrives → `read ECONNRESET` with no listener.
2. **Malformed header, then reset** — a non-`PROXY` line makes the code write a
   diagnostic (`socket.end('* BAD …')` / `'-ERR …'`) and `return` without ever
   invoking the callback; writing to the already-broken socket produces a late
   `write ECONNRESET`/`EPIPE` *after* any naive single error handler would have
   been removed. 

## The fix

Port the guarded `_handleProxy` shape already used in
[`smtp-server`](https://github.com/nodemailer/smtp-server) 3.19.3 (bundled here
as a dependency) into both servers:

- Funnel every exit from the PROXY window through a single-fire `done()` gate.
- Attach `error` + `close` + `end` handlers while waiting, so reset, close, and
  half-close all resolve cleanly (half-close no longer lingers until the socket
  timeout either).
- On the failure path, attach a no-op `error` handler **before** removing our
  own, so a late in-flight error (e.g. the `EPIPE`/`ECONNRESET` from flushing
  the `* BAD`/`-ERR` diagnostic) cannot surface as an uncaughtException.
- Route the malformed-header rejection through `done(err)` as well.
- Both `net.createServer` callers now `return` on error instead of proceeding to
  `connect()` on a dead socket.

## Tests

New regression tests for both servers:

- `imap-core/test/proxy-socket-error-test.js` — reset during the header wait, a
  malformed header followed by a reset, and a valid header that still parses
  (the greeting reports the proxied address).
- `test/pop3-proxy-socket-error-test.js` — reset during the header wait, and a
  valid header that still greets.

They are verified to catch the regression: against the unfixed code the three
crash-survival tests fail with exactly `read ECONNRESET` (×2) and
`write ECONNRESET`; with the fix all pass. Full suite green (`imap`, `pop3`,
unit, `api`) and ESLint clean.

## Reproduce

```js
const net = require('net');
const { IMAPServer } = require('./imap-core/lib/imap-server');

const server = new IMAPServer({ useProxy: ['*'] });
server.on('error', () => {}); // server-level guard
server.listen(0, '127.0.0.1', () => {
    const port = server.server.address().port;
    const client = net.connect(port, '127.0.0.1', () => client.resetAndDestroy());
    client.on('error', () => {});
});
// before the fix: the process exits on an unhandled 'read ECONNRESET'
```

## Scope

- Changed: `imap-core/lib/imap-server.js`, `lib/pop3/server.js`.
- Added: the two test files above.
- Additive and backward-compatible; no API or configuration change.

### DIFF
--- a/imap-core/lib/imap-server.js
+++ b/imap-core/lib/imap-server.js
@@ -58,7 +58,11 @@ class IMAPServer extends EventEmitter {
                 socket.setKeepAlive(true, 5 * 1000);
                 this._handleProxy(socket, (err, socketOptions) => {
                     if (err) {
-                        // ignore, should not happen
+                        // PROXY header could not be read (client dropped the
+                        // connection or sent garbage). The socket is already
+                        // closing/closed and guarded against late errors in
+                        // _handleProxy; just don't proceed to connect().
+                        return;
                     }
                     if (this.options.secured) {
                         return this.connect(socket, socketOptions);
@@ -75,7 +79,9 @@ class IMAPServer extends EventEmitter {
             this.server = net.createServer(this.options, socket =>
                 this._handleProxy(socket, (err, socketOptions) => {
                     if (err) {
-                        // ignore, should not happen
+                        // socket already closing/closed and guarded in
+                        // _handleProxy; just don't proceed to connect()
+                        return;
                     }
                     socket.setKeepAlive(true, 5 * 1000);
                     this.connect(socket, socketOptions);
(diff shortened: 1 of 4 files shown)


NickOvt (changes requested):
- [P1] Malformed PROXY lines can still crash the process. In imap-core/lib/imap-
    server.js:348 and lib/pop3/server.js:399, a line such as PROXY TCP4
    203.0.113.7\r\n passes the command check, then dereferences params[2].trim().
    I reproduced an uncaught TypeError: Cannot read properties of undefined
    (reading 'trim'). Thus an unauthenticated client can still crash either
    server. Validate all required fields—or catch parsing exceptions—and route
    failures through rejectInvalid().

  - [P2] end does not actually close an allowHalfOpen socket. The new handlers in
    imap-core/lib/imap-server.js:315 and lib/pop3/server.js:366 call done(err),
    but never end or destroy the socket. Because no connection object is created,
    no timeout is installed and the socket is not tracked for shutdown. With
    allowHalfOpen: true, I observed getConnections() === 1, readableEnded ===
    true, and writableEnded === false after the FIN. Repeated half-closes can leak
    connections and prevent server.close() from completing. Close/destroy the
    socket before done(), and add the half-close regression test mentioned in the
    comments.

NickOvt (approved):
LGTM

